PR #16, opened 2024-07-02
ENG-1047: AWS Health notifications in Slack

daabr:
Not tested with real data - see https://docs.aws.amazon.com/health/latest/ug/health-api.html

> You must have a Business, Enterprise On-Ramp, or Enterprise Support plan from [AWS Support](https://aws.amazon.com/premiumsupport/) to use the AWS Health API. If you call the AWS Health API from an AWS account that doesn't have a Business, Enterprise On-Ramp, or Enterprise Support plan, you receive a SubscriptionRequiredException error.

### DIFF
--- a/aws_health_to_slack/program.py
+++ b/aws_health_to_slack/program.py
@@ -12,14 +12,14 @@ import re
 
 import autokitteh
 from autokitteh.aws import boto3_client
-from autokitteh.google import google_sheets_client
+from autokitteh.google import google_id, google_sheets_client
 from autokitteh.slack import slack_client
 
 
 url = os.getenv("GOOGLE_SHEET_URL")
 
 
-def on_every_minute(_):
+def on_schedule(_):
     """Workflow's entry-point."""
     slack_channels = _read_google_sheet()
     events = _aws_health_events()
@@ -39,13 +39,8 @@ def on_every_minute(_):
 @autokitteh.activity
 def _read_google_sheet() -> dict[str, str]:
     """Read mapping of project tags to Slack channels from Google Sheet."""
-    match = re.match(r"(.*/d/)?([\w-]{20,})", url)
-    if not match:
-        raise ValueError(f"Error: invalid Google Sheet ID in URL {url!r}")
-
-    id = match.group(2)
     sheets = google_sheets_client("google_sheets_connection").spreadsheets().values()
-    rows = sheets.get(spreadsheetId=id, range="A:B").execute().get("values", [])
+    rows = sheets.get(spreadsheetId=google_id(url), range="A:B").execute().get("values", [])
     return {row[0].strip(): row[1].strip() for row in rows}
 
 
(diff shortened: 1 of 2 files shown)


aws_health_to_slack/program.py
```diff
@@ -0,0 +1,115 @@
...
+
+        channel = slack_channels.get(project)
+        affecting_events = [events_by_arn[arn] for arn in entity["eventArns"]]
+        _post_slack_message(project, channel, entity, affecting_events)
+
+
+@autokitteh.activity
+def _read_google_sheet() -> dict[str, str]:
+    """Read mapping of project tags to Slack channels from Google Sheet."""
+    match = re.match(r"(.*/d/)?([\w-]{20,})", url)
```
pashafateev, line 42: I don't have a specific idea in mind, but thinking about the general user and whether or not they're going to understand regular expressions. It might be intimidating.

aws_health_to_slack/program.py
```diff
@@ -0,0 +1,115 @@
...
+    """List all recent AWS Health events.
+
+    API Reference:
+    https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/health/client/describe_events.html
+    """
+    try:
+        aws = boto3_client("aws_connection", "health")
+        dt = datetime.now(UTC) - timedelta(minutes=1)
+        filter = {"lastUpdatedTime": [{"from": dt}]}
+        # Possible alternative: describe_events_for_organization.
```
pashafateev, line 63: I'm not sure I understand what this comment's intended purpose is. Is this a comment for reviewers to consider a different name for the function below? Or is it meant for the user?

pashafateev (commented):
A few minor comments

aws_health_to_slack/program.py
daabr: `describe_events` and `describe_events_for_organization` are similar-but-different functions in `boto3` (the AWS SDK).

The project is currently using `describe_events` to fetch events for a single AWS account, but with some more setup code we could easily replace it with `describe_events_for_organization` to fetch events for **multiple** AWS accounts.

This would be more useful for large-scale users (i.e. mid-to-large companies), but also more complicated and less appealing for people who want to read the code.

So I guess this is somewhere between a self-reminder and a suggestion for "the interested reader", but not an actual TODO.

Do you think moving this comment to the function's docstring as a clear full sentence would be better?

aws_health_to_slack/program.py
pashafateev: I think it's a great idea!

aws_health_to_slack/program.py
pashafateev: For an inexperienced person yes. I think a complete sentence would make it much more clear.
I think that comment would only make sense to someone who knows those two functions. Which may be ok?

aws_health_to_slack/program.py
```diff
@@ -0,0 +1,115 @@
...
+import autokitteh
+from autokitteh.aws import boto3_client
+from autokitteh.google import google_sheets_client
+from autokitteh.slack import slack_client
+
+
+url = os.getenv("GOOGLE_SHEET_URL")
+
+
+def on_every_minute(_):
```
itayd, line 22: i would change this to `on_schedule`, you might change the  manifest and then need to change this.

aws_health_to_slack/program.py
```diff
@@ -0,0 +1,115 @@
...
+@autokitteh.activity
+def _aws_health_events() -> list[dict]:
+    """List all recent AWS Health events.
+
+    API Reference:
+    https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/health/client/describe_events.html
+    """
+    try:
+        aws = boto3_client("aws_connection", "health")
+        dt = datetime.now(UTC) - timedelta(minutes=1)
```
itayd, line 61: make `minutes` configurable.

aws_health_to_slack/program.py
```diff
@@ -0,0 +1,115 @@
...
+        entities = resp.get("entities", [])
+
+        nextToken = resp.get("nextToken")
+        while nextToken:
+            resp = aws.describe_affected_entities(eventArn=arns, nextToken=nextToken)
+            entities += resp.get("entities", [])
+            nextToken = resp.get("nextToken")
+
+        return entities
+    except Exception as e:
```
itayd, line 100: not sure catching `Exception` is a good idea. Isn't there anything more specific?

daabr: I know, catching `Exception` is an antipattern, but I don't have concrete data yet. Added a TODO comment.

aws_health_to_slack/program.py
daabr: Done

aws_health_to_slack/program.py
daabr: See https://github.com/autokitteh/autokitteh/pull/458

pashafateev (approved):

PR #29, opened 2024-07-24
ENG-1247 Add break glass demo with Jira and Slack

pashafateev:


break_glass/program.py
```diff
@@ -0,0 +1,119 @@
...
+    slack.chat_postMessage(channel=APPROVER, blocks=blocks)
+
+
+def parse_event_data(event):
+    form_data = event.data["view"]["state"]["values"]
+    reason = form_data["block_reason"]["reason"]["value"]
+    issue_key = form_data["block_issue_key"]["issue_key"]["value"]
+    base_url = os.getenv("jira_connection__AccessURL")
+    requester_id = event.data["user"]["id"]
+    requester_name = slack.users_info(user=requester_id)["user"]["name"]
```
daabr, line 103: No need for this call - you can just return the ID, and `<@U1234>` would still be rendered by Slack as the user's display name (see https://api.slack.com/reference/surfaces/formatting#mentioning-users).

The ID is also guaranteed to remain unique forever, whereas the username is guaranteed to remain unique only while the referenced user exists and isn't replaced by a different one.

break_glass/program.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+APPROVER = "sre-team"
+jira = atlassian_jira_client("jira_connection")
+slack = slack_client("slack_connection")
+
+
+def on_slack_slash_command(event):
+    """Sends a form to request approval for a ticket."""
+    trigger_id = event.data["trigger_id"]
```
daabr, line 40: As discussed offline, let's use a "hot word" to prevent triggering of multiple projects that use Slack slash commands, e.g. `/ak break-glass`.

i.e. in this function check that the text of the command (after `.lower().strip()`) matches the regular expression `r"^break.?glass$" - and if not, return and do nothing.

So all the following will work:
* breakglass
* Break-Glass
* BREAK GLASS
* etc.

break_glass/program.py
```diff
@@ -0,0 +1,119 @@
...
+    request_modal = Path("request_modal.json.txt").read_text()
+    slack.views_open(trigger_id=trigger_id, view=request_modal)
+
+
+@autokitteh.activity
+def on_form_submit(event):
+    reason, issue_key, base_url, requester_id, requester_name = parse_event_data(event)
+
+    if not check_issue_exists(issue_key):
+        message = f"Ticket {issue_key} does not exist. Please try again."
```
daabr, line 50:
```suggestion
        message = f"Ticket `{issue_key}` does not exist. Please try again."
```

break_glass/program.py
```diff
@@ -0,0 +1,119 @@
...
+        slack.chat_postMessage(channel=requester_id, text=message)
+        return
+
+    email = slack.users_info(user=requester_id)["user"]["profile"]["email"]
+    if not validate_requester(issue_key, email):
+        message = f"You are not the assignee for ticket {issue_key}. Please try again."
```
daabr, line 56: Please linkify `{issue_key}`, now that it's guaranteed to exist.

Also, replace "for ticket" with "in"

break_glass/program.py
```diff
@@ -0,0 +1,119 @@
...
+import os
+from pathlib import Path
+
+import autokitteh
+from autokitteh.atlassian import atlassian_jira_client
+from autokitteh.slack import slack_client
+from requests.exceptions import HTTPError
+
+
+APPROVER = "sre-team"
```
daabr, line 33: We want this to be configurable. Keep this global value, but initialize it with `os.getenv` (i.e. read a project var)

I'd also rename this to `APPROVAL_CHANNEL` - but only if it doesn't mess up the code by being longer.

break_glass/request_modal.json.txt
```diff
@@ -0,0 +1,47 @@
...
+    {
+      "type": "input",
+      "block_id": "block_issue_key",
+      "element": {
+        "type": "plain_text_input",
+        "action_id": "issue_key"
+      },
+      "label": {
+        "type": "plain_text",
+        "text": "Ticket Reference ID",
```
daabr, line 28: Nit: "Reference Ticket ID" instead of "Ticket Reference ID".

The ticket is the reference proving that requester has a valid need for access, not the ID of the ticket.

pashafateev: Done

daabr (approved):

PR #34, opened 2024-08-01
add `required` field to varargs where it makes sense

itayd:
see https://github.com/autokitteh/autokitteh/pull/537

should we add a `description` field for each var?


daabr (commented):
I think this PR is obsolete, based on https://github.com/autokitteh/autokitteh/pull/543.

Also, FYI, I created a ticket for this: https://linear.app/autokitteh/issue/ENG-1367/update-var-values-in-samples-kittehub-internal-demos

Lastly, adding an optional description field to vars seems like a nice idea!

PR #38, opened 2024-08-07
fix(eng-1393): handle new emails only and improve label creation

pashafateev:
I am aware that it makes the code longer, but the other version is too error-prone and I could not get it to work.

- Ensure only new emails received after the script starts running are processed and sent to Slack.
- Implement pagination handling to correctly retrieve all emails from Gmail API.
- Improve error handling and logging for label creation to ensure valid label IDs are used in email modification requests.
- Fix issue where total message count included draft emails, causing incorrect triggers.


categorize_notify/program.py
```diff
@@ -10,60 +10,86 @@
...
 import time
+from datetime import datetime, timezone
 
 import autokitteh
 from autokitteh import google, openai, slack
 
 
 SLACK_CHANNELS = ["demos", "engineering", "ui"]
 
+processed_message_ids = set()
```
itayd, line 21: why are these global?

categorize_notify/program.py
```diff
@@ -10,60 +10,86 @@
...
+start_time = datetime.now(timezone.utc)
+
 
 def on_http_get(event):
-    total_messages = None
     while True:
-        total_messages = _poll_inbox(total_messages)
+        _poll_inbox()
         time.sleep(10)
```
itayd, line 28: i would do the duration configurable via vars

categorize_notify/program.py
```diff
@@ -10,60 +10,86 @@
...
 
 
 @autokitteh.activity
```
itayd, line 31: is this really necessary?

categorize_notify/program.py
```diff
@@ -10,60 +10,86 @@
...
-        new_email_count = curr_total - prev_total
-        message_ids = _get_latest_message_ids(gmail, new_email_count)
-        for message_id in message_ids:
-            _process_email(gmail, message_id)
+    token = None
+    current_message_ids = set()
+
+    while True:
+        query = "in:inbox -in:drafts"
+        if token:
```
itayd, line 39: my guess is that only the inside of the loop needs to be an activity, not the entire loop

categorize_notify/program.py
```diff
@@ -10,60 +10,86 @@
...
+
+    processed_message_ids.update(new_message_ids)
 
 
 def _process_email(gmail, message_id: str):
     message = gmail.messages().get(userId="me", id=message_id).execute()
+    email_timestamp = int(message["internalDate"]) / 1000
+    email_datetime = datetime.fromtimestamp(email_timestamp, timezone.utc)
+
+    if email_datetime < start_time:
```
itayd, line 64: no chance of missing stuff here with same start_time but that were not read yet?

categorize_notify/program.py
```diff
@@ -10,60 +10,86 @@
...
     email_content = _parse_email(message)
+
     if email_content:
         channel = _categorize_email(email_content)
+
         if channel:
             client = slack.slack_client("my_slack")
-            client.chat_postMessage(channel=channel, text=email_content)
+            client.chat_postMessage(
+                channel=channel, text=email_content or "Empty email"
```
itayd, line 75: nit: i would split further into two lines

categorize_notify/program.py
```diff
@@ -10,60 +10,86 @@
...
+            body = {"addLabelIds": [label_id]}
+            gmail.messages().modify(userId="me", id=message_id, body=body).execute()
 
 
 def _parse_email(message: dict):
     payload = message["payload"]
-    for part in payload.get("parts", []):
-        if part["mimeType"] == "text/plain":
-            return base64.urlsafe_b64decode(part["body"]["data"]).decode("utf-8")
+    parts = payload.get("parts", [])
```
itayd, line 87:
```suggestion
    payload = message["payload"].get("parts", [])
```

categorize_notify/program.py
pashafateev: Done

categorize_notify/program.py
pashafateev: but I don't want the entire "parts", I just want the text
https://googleapis.github.io/google-api-python-client/docs/dyn/gmail_v1.users.messages.html#list

categorize_notify/program.py
```diff
@@ -9,64 +9,106 @@
...
+import os
 import time
 
 import autokitteh
 from autokitteh import google, openai, slack
 
 
+POLL_INTERVAL = os.getenv("POLL_INTERVAL")
 SLACK_CHANNELS = ["demos", "engineering", "ui"]
+gmail = google.gmail_client("my_gmail").users()
```
daabr, line 22:
```suggestion
SLACK_CHANNELS = ["demos", "engineering", "ui"]


gmail = google.gmail_client("my_gmail").users()
```

categorize_notify/program.py
```diff
@@ -9,64 +9,106 @@
...
 import base64
+from datetime import datetime, timezone
+import os
 import time
 
 import autokitteh
 from autokitteh import google, openai, slack
 
 
+POLL_INTERVAL = os.getenv("POLL_INTERVAL")
```
daabr, line 20:
```suggestion
POLL_INTERVAL = float(os.getenv("POLL_INTERVAL"))
```

categorize_notify/program.py
```diff
@@ -9,64 +9,106 @@
...
+start_time = datetime.now(timezone.utc).timestamp()
 
 
 def on_http_get(event):
-    total_messages = None
     while True:
-        total_messages = _poll_inbox(total_messages)
-        time.sleep(10)
+        _poll_inbox()
+        time.sleep(float(POLL_INTERVAL))
```
daabr, line 30:
```suggestion
        time.sleep(POLL_INTERVAL)
```

categorize_notify/program.py
```diff
@@ -9,64 +9,106 @@
...
+
+def _poll_inbox():
+    token = None
+    current_message_ids = set()
+
+    while True:
+        results = get_new_inbox_messages(token)
+
+        current_message_ids.update({msg["id"] for msg in results.get("messages", [])})
+
```
daabr, line 41:
```suggestion
        current_message_ids.update({msg["id"] for msg in results.get("messages", [])})
```

categorize_notify/program.py
daabr: I tend to prefer the global approach for simplicity - the local approach requires passing it, understanding Python mutability of passed sets, etc.

categorize_notify/program.py
```diff
@@ -9,64 +9,106 @@
...
+        token = results.get("nextPageToken")
+        if not token:
+            break
+
+    new_message_ids = current_message_ids - processed_message_ids
+
+    for message_id in new_message_ids:
+        _process_email(message_id, start_time)
+
+    processed_message_ids.update(new_message_ids)
```
daabr, line 51: Are you **sure** it works without specifying that `processed_message_ids` is `global`? It should be required, but even if not it's a good self-documenting clarification.

```suggestion
    global processed_message_ids
    processed_message_ids.update(new_message_ids)
```

categorize_notify/program.py
```diff
@@ -9,64 +9,106 @@
...
-        message_ids = _get_latest_message_ids(gmail, new_email_count)
-        for message_id in message_ids:
-            _process_email(gmail, message_id)
+def get_new_inbox_messages(token):
+    query = "in:inbox -in:drafts"
+    if token:
+        gmail_list = gmail.messages().list(userId="me", q=query, pageToken=token)
+        results = gmail_list.execute()
+    else:
+        results = gmail.messages().list(userId="me", q=query).execute()
```
daabr, line 61: Please test if passing the tokens `""` and `None` has the same effect like not specifying a token at all - if so, remove the `if` condition and the `else` branch.

categorize_notify/program.py
```diff
@@ -9,64 +9,106 @@
...
-        label_id = _get_label_id(gmail, channel) or _create_label(gmail, channel)
-        body = {"addLabelIds": [label_id]}
-        gmail.messages().modify(userId="me", id=message_id, body=body).execute()
+        label_id = _get_label_id(channel) or _create_label(channel)
+        if label_id:
+            body = {"addLabelIds": [label_id]}
+            apply_label_to_message(message_id, body)
+
+
+@autokitteh.activity
```
daabr, line 89: Please recheck if the decorator is still needed after  https://github.com/autokitteh/autokitteh/pull/540

categorize_notify/program.py
```diff
@@ -76,12 +118,15 @@ def _create_label(gmail, label_name: str) -> str:
         "messageListVisibility": "show",
         "name": label_name,
     }
-    created_label = gmail.labels().create(userId="me", body=label).execute()
-    print(f"Label created: {created_label['name']}")
-    return created_label["id"]
+    try:
+        created_label = gmail.labels().create(userId="me", body=label).execute()
+    except Exception as e:
+        print(f"Error creating label: {e}")
```
daabr, line 124: I think this `try-except` block is unnecessary: now that exceptions dont' get swallowed by the runtime, we should keep raising this as an error, not just report it in the session log. Unless your think otherwise?

daabr (approved):

categorize_notify/program.py
pashafateev: >In Python, the global keyword allows you to change a variable value outside of its current scope. It is used to make changes to a global variable from a local location. The global keyword is only required for altering the variable value and not for publishing or accessing it. [source](https://www.toppr.com/guides/python-guide/references/methods-and-functions/global-keyword/python-global-keyword-with-examples/)

>Minimize Use: Only use global variables when necessary. [source](https://lifewithdata.com/2023/10/20/python-global-keyword/)

Because the set isn't being reassigned to another set or something else entirely, global is not required. I'm not 100% but I don't think it would be considered best practice.

categorize_notify/program.py
pashafateev: I thought about that too and I don't understand why but when I do it the way you're suggesting I get this error:
[error.txt](https://github.com/user-attachments/files/16593270/error.txt)

I ran it twice each way and got consistent results. Which I understand doesn't prove anything. I'll try a few more times later.

pashafateev: After talking with Itay, I decided to remove pagination entirely. I get the first 100 results by default and they seem to be in reverse chronological order although I could not find any guarantees from Google.

I think this is still better than paginating the entire inbox.

categorize_notify/program.py
pashafateev: It is only needed for the chatgpt call. Removed everywhere else.

categorize_notify/program.py
pashafateev: You're right. I thought it was needed, but it was a different issue that I fixed. `label_name` was sometimes `None` and that was causing an error.

PR #60, opened 2024-09-08
Update 3 projects to use new webhooks

daabr:
Note to reviewers: due to the amount and complexity of changes, it's better to review the final result instead of side-by-side with the previous state. Otherwise you're comparing apples to oranges.

- Quickstart
- Google Forms to Jira (rewritten to use events instead of polling)
- Webhook to Jira (nee "create Jira issue")

Also includes readme updates.

Refs: ENG-1509


google_forms_to_jira/README.md
```diff
@@ -1,85 +1,81 @@
...
 
-1. **Trigger**: The workflow is triggered by an HTTP GET request.
-2. **Poll Forms**: The program polls the specified Google Form for new responses.
-3. **Create Jira Issue**: For each new response, the program creates a Jira issue with the response data.
-
-## Installation and Usage 
-
-- [Install AutoKitteh](https://docs.autokitteh.com/get_started/install)
+1. The workflow is triggered by response events from Google Forms, for a
+   specific form ID
```
pashafateev, line 8:
```suggestion
1. The workflow is triggered by response events from a specific Google Form, identified by its unique form ID
```

google_forms_to_jira/program.py
```diff
@@ -1,59 +1,69 @@
...
 import os
-import time
 
-import autokitteh
 from autokitteh import google
 from autokitteh.atlassian import jira_client
 
 
-POLL_INTERVAL = os.getenv("POLL_INTERVAL")
+JIRA_PROJECT_KEY = os.getenv("JIRA_PROJECT_KEY")
```
pashafateev, line 16: Is this the convention we're using?

**Connection-related strings (e.g., "forms_conn"):** If a connection string is only used a few times in the code, we prefer to rewrite the string directly rather than define a constant. This avoids cluttering the code with unnecessary constants for values that aren't reused extensively.

**Everything else**: For other values that are used frequently or could benefit from a single source of truth (e.g., form IDs, API keys, or frequently referenced text), we use constants. This ensures consistency, reduces the risk of typos, and makes the code easier to update when the value needs to change."

google_forms_to_jira/program.py
```diff
@@ -1,59 +1,69 @@
...
+
+    answers = _summarize_form_response(event.data.response.answers, questions)
+
+    # Check if a Jira issue already exists for this response (i.e. response edited).
+    query = f"project = {JIRA_PROJECT_KEY} AND description ~ {response_id}"
+    issues = jira.jql(query + " ORDER BY created DESC")
+    if issues.get("total", 0) == 0:
+        _create_jira_issue(answers, response_id)
+    else:
+        _update_jira_issue(issues["issues"][0], answers, response_id)
```
pashafateev, line 36:
```suggestion
def on_form_response(event):
    data = event.data
    print("Form response submitted:", data)
    response_id = data.response.response_id

    forms = google.google_forms_client("forms_conn").forms()
    questions = forms.get(formId=data.form_id).execute().get("items", [])

    answers = _summarize_form_response(data.response.answers, questions)

    # Check if a Jira issue already exists for this response (i.e. response edited).
    query = f"project = {JIRA_PROJECT_KEY} AND description ~ {response_id}"
    issues = jira.jql(query + " ORDER BY created DESC")
    if issues.get("total", 0) == 0:
        _create_jira_issue(answers, response_id)
    else:
        _update_jira_issue(issues["issues"][0], answers, response_id)
```

google_forms_to_jira/program.py
```diff
@@ -1,59 +1,69 @@
...
+
+jira = jira_client("jira_conn")
+
+
+def on_form_response(event):
+    print("Form response submitted:", event.data)
+    response_id = event.data.response.response_id
+
+    forms = google.google_forms_client("forms_conn").forms()
+    questions = forms.get(formId=event.data.form_id).execute().get("items", [])
```
pashafateev, line 26: this is only loosely related to this PR, but what if we abstracted away these Google functions? They look unnecessarily complicated. so all the user has to call is `forms.get(formId)`

pashafateev (approved):

google_forms_to_jira/program.py
daabr: I get your point, but Google's Python libraries use what's known as a "builder pattern" for their APIs. In our specific samples you don't see the benefit in that design choice, so it looks like unnecessary complexity, but if we were to wrap it as you suggest we would limit the functionality of the library, specifically we'd hide the "customizability" of Google's functions.

Also, writing these wrappers would be like our Go wrappers for Starlark: it would be custom code that people are not familiar with, but we'd have to invest a lot of work into creating and maintaining it.

google_forms_to_jira/program.py
daabr: Yes. I'd also say that when a const value is defined using some action (e.g. reading an env var, initializing an object) then it makes more sense to do it only in one place.

google_forms_to_jira/program.py
daabr: That's a topic to discuss with Miki. I'd rather save users the need to set `data = event.data`.

PR #64, opened 2024-09-11
feat: GitHub Marketplace to Slack

daabr:
Originally created (as a draft PR) for AutoKitteh's own GitHub Marketplace webhook. Updated it now so it's ready for public usage.

See readme draft here: https://github.com/autokitteh/kittehub/blob/55d2f95d81c7acedc54bcef0cf060d2ba959fcc0/github_marketplace_to_slack/README.md

### DIFF
--- a/github_marketplace_to_slack/README.md
+++ b/github_marketplace_to_slack/README.md
@@ -1,0 +1,90 @@
+# Webhook to Jira
+
+This project receives
+[webhook notifications from the GitHub Marketplace](https://docs.github.com/en/apps/github-marketplace/listing-an-app-on-github-marketplace/configuring-a-webhook-to-notify-you-of-plan-changes),
+when changes to customer account plans occur, and posts them to a Slack
+channel.
+
+This allows you to handle
+[marketplace_purchase](https://docs.github.com/en/apps/github-marketplace/using-the-github-marketplace-api-in-your-app/webhook-events-for-the-github-marketplace-api)
+events in your GitHub app.
+
+## API Documentation
+
+GitHub:
+
+- [Configuring a webhook to notify you of plan changes](https://docs.github.com/en/apps/github-marketplace/listing-an-app-on-github-marketplace/configuring-a-webhook-to-notify-you-of-plan-changes)
+
+HTTP:
+
+- https://docs.autokitteh.com/integrations/http/events
+
+Slack:
+
+- https://docs.autokitteh.com/integrations/slack/python
+
+## Setup Instructions
+
+1. Install and start a
+   [self-hosted AutoKitteh server](https://docs.autokitteh.com/get_started/quickstart),
+   or use AutoKitteh Cloud
+
+2. Optional for self-hosted servers (preconfigured in AutoKitteh Cloud): \
+
+   - [enable Slack connections to use an OAuth v2 app](https://docs.autokitteh.com/integrations/slack/config)
+
+3. Run this command to clone the Kittehub repository, which contains this
+   project:
+
+   ```shell
+   git clone https://github.com/autokitteh/kittehub.git
+   ```
+
+4. Set these variables in this project's [autokitteh.yaml](./autokitteh.yaml)
+   manifest file:
+
+   - `GITHUB_WEBHOOK_SECRET`
+   - `SLACK_CHANNEL_NAME_OR_ID`
+
+5. Run this command to deploy this project's manifest file:
+
+   ```shell
+   ak deploy --manifest kittehub/github_marketplace_to_slack/autokitteh.yaml
+   ```
+
+6. Look for the following line in the output of the `ak deploy` command, and
+   copy the URL path for later:
+
+   ```
+   [!!!!] trigger "webhook_notification" created, webhook path is "/webhooks/..."
+   ```
+
+> [!TIP]
+> If you don't see the output of `ak deploy` anymore, you can run this command
+> instead, and use the webhook slug from the output:
+>
+> ```shell
+> ak trigger get webhook_notification --project github_marketplace_to_slack -J
+> ```
+
+7. Initialize this project's Slack connection with an OAuth v2 app (based on
+   step 2), or a Socket Mode app
+
+> [!TIP]
+> The exact CLI command to do so (`ak connection init ...`) will appear in the
+> output of the `ak deploy` command from step 5 when you create the project on
+> the server, i.e. when you run that command for the first time.
+
+## Usage Instructions
+
+1. Configure the GitHub app's webhook in the GitHub Marketplace:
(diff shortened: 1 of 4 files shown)


github_marketplace_to_slack/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+Configure the GitHub app's webhook in the GitHub Marketplace:
+`https://github.com/marketplace/YOUR-APP-NAME/hook`
+
+- **Payload URL:** _(you will set this later, see steps 3 and 6 in the_
+  _[Cloud Usage](#cloud-usage) section below)_
+- **Content type:** `application/json`
+- **Secret:** a random string of text with high entropy, save for later
+
+Don't click the "Create/Update webhook" button yet.
```
pashafateev:
```suggestion
> [!IMPORTANT]
> Don't click the "Create/Update webhook" button yet.
```

github_marketplace_to_slack/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Cloud Usage
+
+1. Import/upload the project
```
pashafateev: Does this need to happen in a separate window?

github_marketplace_to_slack/README.md
```diff
@@ -0,0 +1,83 @@
...
+2. Initialize your connections
+3. Edit the trigger:
```
pashafateev: as discussed in integration sync. No need to actually edit.

github_marketplace_to_slack/README.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+> [!TIP]
+> If you don't see the output of `ak deploy` anymore, you can run this command
+> instead, and use the `webhook_slug` from the output:
+>
+> ```shell
+> ak trigger get webhook_notification --project github_marketplace_to_slack -J
+> ```
+
```
pashafateev: I’d avoid including this in our documentation. It feels similar to the boilerplate content we replaced with a link in the self-hosted section. Instead, I’d suggest creating a separate Docusaurus page and linking to it. This approach avoids the question of which READMEs should include CLI instructions. If we add them to all, it leads to clutter. If we include them per project, it reopens discussions about consistency.

I'm happy to add the page if you agree with the idea. @mario99logic just making you aware

github_marketplace_to_slack/program.py
```diff
@@ -0,0 +1,61 @@
...
+import hmac
+import json
+import os
+
+import autokitteh
+from autokitteh.slack import slack_client
+
+
+def on_webhook_notification(event):
+    """Handle GitHub Marketplace webhook notifications."""
```
pashafateev: Are we intentionally choosing not to include the event parameter in the docstring’s parameter list? If so, what’s the reasoning behind that decision?

github_marketplace_to_slack/README.md
daabr: discussed offline

github_marketplace_to_slack/README.md
daabr: done

github_marketplace_to_slack/README.md
daabr: https://github.com/autokitteh/docs.autokitteh.com/pull/147

github_marketplace_to_slack/program.py
daabr: fixed

github_marketplace_to_slack/README.md
```diff
@@ -0,0 +1,66 @@
...
+- **Payload URL:** _(you will set this later, based on step 2 in the_
+  _[Cloud Usage](#cloud-usage) section below)_
+- **Content type:** `application/json`
+- **Secret:** a random string of text with high entropy _(save it for later,_
+  _in step 3 in the [Cloud Usage](#cloud-usage) section below)_
+
+> [!IMPORTANT]
+> Do not click the "Create/Update webhook" button, and do not close this
+> browser tab yet - you will need to get back here later, in step 5 in the
+> [Cloud Usage](#cloud-usage) section below
```
pashafateev:
```suggestion
> Do not click the "Create/Update webhook" button and keep this
> browser tab open. You will need to return here later, in step 5 of the
> [Cloud Usage](#cloud-usage) section below.
```

pashafateev (approved):

github_marketplace_to_slack/program.py
```diff
@@ -0,0 +1,65 @@
+"""Forward GitHub Marketplace webhook notifications to a Slack channel.
+
+GitHub API documentation:
+- https://docs.github.com/en/apps/github-marketplace/listing-an-app-on-github-marketplace/configuring-a-webhook-to-notify-you-of-plan-changes
+- https://docs.github.com/en/webhooks/webhook-events-and-payloads#marketplace_purchase
+- https://docs.github.com/en/webhooks/webhook-events-and-payloads#ping
```
mario99logic, line 6: Why add the url here instead of the README

github_marketplace_to_slack/program.py
```diff
@@ -0,0 +1,65 @@
...
+    secret = os.getenv("GITHUB_WEBHOOK_SECRET", "")
+    signature = headers.get("X-Hub-Signature-256", "")
+    _verify_signature(event.data.body.bytes, secret, signature)
+
+    # https://docs.github.com/en/webhooks/webhook-events-and-payloads#delivery-headers
+    msg = f"*GitHub Marketplace event:* `{headers.get('X-Github-Event', '???')}`\n"
+    msg += f"(Resource ID `{headers.get('X-Github-Hook-Installation-Target-Id')}`, "
+    msg += f"webhook ID `{headers.get('X-Github-Hook-Id')}`, "
+    msg += f"event ID `{headers.get('X-Github-Delivery')}`)\n"
+
```
mario99logic, line 35: Why the extra space

github_marketplace_to_slack/program.py
daabr: The first link is in the readme too. The other 2 a relevant for implementers - people who want to modify this code, not just run it, so according to the proximity principle this is the more appropriate place for them. But it's not a bad idea to add them in the readme too.

github_marketplace_to_slack/program.py
daabr: I'm appending to the same string, so the spaces ensure the final string will be "one two three" instead of "onetwothree".

mario99logic: no I mean the extra space between line 34 and line 36

daabr: oh, because they are separate blocks with different functionalities: lines 30-34 examine HTTP headers, lines 36-38 examine the HTTP payload.

PR #66, opened 2024-09-11
feat(): Add Slack-Discord message mirroring project

pashafateev:
Refs: eng-1510

### DIFF
--- a/slack_discord_sync/program.py
+++ b/slack_discord_sync/program.py
@@ -1,0 +1,74 @@
+"""
+This script mirrors messages between Slack and Discord channels using
+autokitteh's Slack and Discord clients.
+
+Imports:
+- `autokitteh.discord as ak_discord`: Wrapper around the Discord API
+  client to simplify authentication and secret management.
+- `slack`: Module from autokitteh that helps interact with the Slack API.
+- `discord`: Necessary import for certain configurations, such as
+  enabling message intents, and handling exceptions as outlined in the
+  official Discord.py quickstart documentation
+  (https://discordpy.readthedocs.io/en/stable/quickstart.html?highlight=
+  send%20message#quickstart).
+
+Global Variables:
+- `intents`: Discord intents used to enable the bot to read message
+  content.
+- `client`: The autokitteh Discord client initialized with the necessary
+  intents and connection settings.
+- `slack_api`: The autokitteh Slack client used to post messages to Slack.
+- `slack_message`: Stores the latest message received from Slack to be
+  posted to Discord.
+
+- `on_ready()`: An asynchronous event triggered when the Discord bot
+  successfully connects. It fetches the Discord channel by ID and sends
+  the latest message received from Slack to the channel, then closes the
+  client connection.
+
+Note:
+The `discord` import is crucial for enabling specific Discord API
+configurations, such as intents and error handling, but all functional
+API calls are made through the autokitteh `ak_discord` wrapper, which
+streamlines authentication and secret management.
+"""
+
+import os
+
+from autokitteh import discord as ak_discord
+from autokitteh import slack
+import discord
+
+
+DISCORD_CHANNEL_ID = int(os.getenv("DISCORD_CHANNEL_ID"))
+SLACK_CHANNEL_ID = os.getenv("SLACK_CHANNEL_ID")
+
+intents = discord.Intents.default()
+intents.message_content = True
+client = ak_discord.discord_client("discord_conn", intents)
+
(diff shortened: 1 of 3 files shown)


slack_discord_sync/program.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def on_discord_message(event):
+    slack_api.chat_postMessage(channel=SLACK_CHANNEL_ID, text=event.data["content"])
+
+
+def on_slack_message(event):
+    global slack_message
+    slack_message = event.data["text"]
+    client.run(ak_discord.bot_token("discord_conn"))
```
daabr, line 61: So for every new Slack message you create another Discord client? Am I understanding this correctly?

slack_discord_sync/program.py
```diff
@@ -0,0 +1,74 @@
+"""
+This script mirrors messages between Slack and Discord channels using
+autokitteh's Slack and Discord clients.
```
daabr:
```suggestion
AutoKitteh's Slack and Discord clients.
```

slack_discord_sync/program.py
```diff
@@ -0,0 +1,74 @@
...
+
+Imports:
```
daabr: I think the imports section here is an overkill - we don't explain about `autokitteh.*` imports anywhere else, and `discord` is self-explanatory. But keep the link to Discord's documentation - see https://github.com/autokitteh/kittehub/tree/main/google_forms_to_jira for example on how I specify this in the code and in the readme.

slack_discord_sync/program.py
```diff
@@ -0,0 +1,74 @@
...
+- `autokitteh.discord as ak_discord`: Wrapper around the Discord API
+  client to simplify authentication and secret management.
+- `slack`: Module from autokitteh that helps interact with the Slack API.
+- `discord`: Necessary import for certain configurations, such as
+  enabling message intents, and handling exceptions as outlined in the
+  official Discord.py quickstart documentation
+  (https://discordpy.readthedocs.io/en/stable/quickstart.html?highlight=
+  send%20message#quickstart).
+
+Global Variables:
```
daabr: Better: document each global var, instead of a "faraway" section here. We may way to tweak some of the names, but that's no important for now.

slack_discord_sync/program.py
```diff
@@ -0,0 +1,74 @@
...
+- `intents`: Discord intents used to enable the bot to read message
+  content.
+- `client`: The autokitteh Discord client initialized with the necessary
+  intents and connection settings.
+- `slack_api`: The autokitteh Slack client used to post messages to Slack.
+- `slack_message`: Stores the latest message received from Slack to be
+  posted to Discord.
+
+- `on_ready()`: An asynchronous event triggered when the Discord bot
```
daabr: This should be a docstring in the function itself.

slack_discord_sync/program.py
```diff
@@ -0,0 +1,74 @@
...
+  successfully connects. It fetches the Discord channel by ID and sends
+  the latest message received from Slack to the channel, then closes the
+  client connection.
+
+Note:
```
daabr, line 8: This is a good note

slack_discord_sync/program.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+DISCORD_CHANNEL_ID = int(os.getenv("DISCORD_CHANNEL_ID"))
+SLACK_CHANNEL_ID = os.getenv("SLACK_CHANNEL_ID")
+
+intents = discord.Intents.default()
+intents.message_content = True
+client = ak_discord.discord_client("discord_conn", intents)
+
+slack_api = slack.slack_client("slack_conn")
```
daabr, line 30: Given that it's used only in 1 place below, it's better to initialize it locally there instead of a global var

slack_discord_sync/program.py
pashafateev: Done

slack_discord_sync/program.py
pashafateev: Done

slack_discord_sync/program.py
pashafateev: Done

slack_discord_sync/program.py
pashafateev: I agree. However, from a visual perspective, it might be better to keep them together and avoid differentiating between the two clients, as if there’s a specific reason one is local and the other is global. That said, it’s not a hill I’m going to die on.

daabr (approved):

PR #72, opened 2024-09-24
fix(samples): switch http to webhooks and bug fixes

pashafateev:
[related PR](https://github.com/autokitteh/autokitteh/pull/690)


github_copilot_seats/triggers.star
haimzlato, line 1: We should convert all .star to Python...

github_copilot_seats/autokitteh.yaml
```diff
@@ -16,14 +16,14 @@ project:
...
     - name: GITHUB_ORG
-      value: "<YOUR_GITHUB_ORG>"
+      value: autokitteh
     - name: IDLE_USAGE_THRESHOLD
       value: "72h"
     - # If not empty, only manage GitHub Copilot subscriptions to the specified users, separated by commas.
       name: LOGINS
       value: ""
     - name: LOG_CHANNEL
-      value: ""
```
haimzlato, line 26: Same here, shouldn't it be empty?

github_copilot_seats/autokitteh.yaml
```diff
@@ -16,14 +16,14 @@ project:
   name: github_copilot_seats
   vars:
     - name: GITHUB_ORG
-      value: "<YOUR_GITHUB_ORG>"
+      value: autokitteh
```
haimzlato: For the templates, we agreed to leave the values empty. @daabr am I wrong?

github_copilot_seats/triggers.star
daabr: I agree with Haim, but let's do this separately from this PR

github_copilot_seats/autokitteh.yaml
daabr: it doesn't matter anymore because we don't warn the user when they forget to specify a value, but I agree that for well-understood values (like a GitHub org name) an empty value is better than `"<DESCRIPTION OF VAR>"` or a specific default which may not be suitable for most users

github_copilot_seats/autokitteh.yaml
daabr: I'm ok with either "" and "log"

github_copilot_seats/autokitteh.yaml
```diff
@@ -35,5 +35,5 @@ project:
       call: triggers.star:on_schedule
     - name: check_now
       connection: myslack
-      event_type: app_mention
-      call: triggers.star:on_slack_app_mention
+      event_type: slash_command
```
daabr, line 38: Personally I also prefer slash commands over app mentions: app mention events are sometimes competing with message events when used in the bot's DM channel, they work best in regular (public or private) channels that the bot has access to.

But using slash commands as triggers instead of other Slack event types is not an official convention that we have.

Do you have a specific reason not to use an app mention? Or was it just convenience?

github_copilot_seats/triggers.star
```diff
@@ -4,15 +4,14 @@ load("seats.star", "prune_idle_seats", "find_idle_seats")
 def on_schedule():
     print(prune_idle_seats())
 
-def on_slack_app_mention(data):
-    parts = data.text.split(" ")
-    if len(parts) < 2:
-        myslack.chat_post_message(data.channel, "unrecorgnized command", thread_ts=data.ts)
+def on_slack_slash_command(data):
```
daabr, line 7: :+1: This new logic for command parsing is simpler

github_copilot_seats/triggers.star
```diff
@@ -4,15 +4,14 @@ load("seats.star", "prune_idle_seats", "find_idle_seats")
...
+    cmd = data.text
+    if cmd not in ["prune-idle-copilot-seats", "find-idle-copilot-seats"]:
+        myslack.chat_post_message(data.channel_id, "unrecognized command")
         return
 
-    cmd = parts[1].strip()
 
-    reply = lambda msg: myslack.chat_post_message(data.channel, msg, thread_ts=data.ts)
+    reply = lambda msg: myslack.chat_post_message(data.channel_id, msg)
```
daabr, line 14: 1. Reminder for when this is is converted to Python: get rid of `lambda`

2. The original code replied in the same thread of the user's app mention - this is not an option for slash commands (I think they're invisible to other users), but also maybe the response should be an ephemeral message instead of a regular one? wdyt?

github_copilot_seats/README.md
```diff
@@ -8,3 +8,6 @@ Users can then optionally ask for their subscription to be reinstated.
 
 - `/ak prune-idle-copilot-seats` invokes the automation immediately.
 - `/ak find-idle-copilot-seats` displays the potentially idle seats.
+
+> [!WARNING]
+> This example currently works only with a [Personal Access Token](https://docs.autokitteh.com/integrations/github/connection/#personal-access-token-pat), specifically a [classic token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-personal-access-token-classic.)
```
daabr:
```suggestion
> This example currently works only with a [Personal Access Token](https://docs.autokitteh.com/integrations/github/connection/#personal-access-token-pat), specifically a [classic token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-personal-access-token-classic).
```

aws_health_to_slack/program.py
```diff
@@ -87,13 +87,18 @@ def _affected_aws_entities(events: list[dict]) -> list[dict]:
...
         aws = boto3_client("aws_connection", "health")
         arns = [event["arn"] for event in events]
+
+        if not arns:
+            return []
+
+        filter = {"eventArns": arns}
         # Possible alternative: describe_affected_entities_for_organization.
-        resp = aws.describe_affected_entities(eventArn=arns)
+        resp = aws.describe_affected_entities(filter=filter)
```
daabr, line 97: Good catch 👍

aws_health_to_slack/program.py
```diff
@@ -87,13 +87,18 @@ def _affected_aws_entities(events: list[dict]) -> list[dict]:
     try:
         aws = boto3_client("aws_connection", "health")
         arns = [event["arn"] for event in events]
+
+        if not arns:
+            return []
```
daabr: I think a clearer place for this check is after line 26 (which is also affected by line 25, but won't fail if `events` is empty.

The hard-to-understand subtlety about this check here is that there's no risk that a (real) event won't contain an "arn" field, but it is possible that `events` is empty, in which case `filter` will not contain a required field (`eventArns`).

pashafateev: Done

github_copilot_seats/README.md
pashafateev: Done

github_copilot_seats/autokitteh.yaml
pashafateev: It was an accident. I was testing and forgot to delete. Done.

github_copilot_seats/autokitteh.yaml
pashafateev: Based on the README, I thought they were supposed to be slash commands.

github_copilot_seats/triggers.star
pashafateev: `chat_post_ephemeral(channel=data.channel_id, text="", user=data.user_id)`
I like this idea, but the function uses kwargs and it requires the slack user id which isn't available in all of the contexts

daabr (approved):

PR #78, opened 2024-10-01
feat: port twilio to python

pashafateev:
Based on our discussion about simplifying the samples, I have removed the Slack integration component.


samples/twilio/program.py
```diff
@@ -0,0 +1,58 @@
...
+
+    The phone number to send the message to must be provided in the query
+    parameter 'to'. The message will be sent both as an SMS and a WhatsApp
+    message to the specified number.
+
+    Args:
+        event (object): An event object containing the request data.
+    """
+    to = event.data["url"]["query"]["to"]
+    try:
```
itayd: this try except is too wide. catch specific exception on smaller scopes.

pashafateev: I removed the try except entirely. I could do a try except around each message sending, but I'm not sure it's that necessary here?

itayd (approved):

PR #80, opened 2024-10-02
feat: remove slack from gmail sample

pashafateev:


samples/google/gmail/program.py
```diff
@@ -9,144 +9,134 @@
...
 
 
-def on_slack_slash_command(event):
-    """Use a Slack slash command to interact with a Gmail mailbox.
+def on_http_get(event):
+    """Handle Gmail interaction via HTTP trigger using query params.
 
-    See: https://api.slack.com/interactivity/slash-commands, and
-    https://api.slack.com/interactivity/handling#message_responses
+    Example URL: "http://localhost:9980/webhooks/<webhook_slug>?cmd=list_drafts"
```
daabr, line 21: This definitely work, nice!

Need to update the readme too.

See step 1 in the usage instructions in the readme of the HTTP sample about documenting how to send HTTP requests with parameters that are not attached to the URL. I think it's cleaner, but it's up to you (the list of commands below is short and to the point, which is also good in comparison to a long description of curl commands).

samples/google/gmail/autokitteh.yaml
```diff
@@ -9,13 +9,10 @@ project:
...
-    - name: slack_conn
-      integration: slack
   triggers:
-    - name: slack_slash_command
-      connection: slack_conn
-      event_type: slash_command
-      call: program.py:on_slack_slash_command
+    - name: on_http_get
+      type: webhook
+      call: program.py:on_http_get
```
daabr:
```suggestion
      type: webhook
      event_type: get
      call: program.py:on_http_get
```

samples/google/gmail/program.py
```diff
@@ -9,144 +9,134 @@
...
-
-    Args:
-        slack_channel: Slack channel name/ID to post debug messages to.
-        _: Unused suffix of the user's Slack command, if any.
+        event: HTTP request event data (contains query parameters).
     """
+    params = event.data.url.query
+    cmd = params.get("cmd")
+
+    if cmd == "get_profile":
```
daabr: Better in this case: Python's "switch" (https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching)

samples/google/gmail/program.py
```diff
@@ -9,144 +9,134 @@
...
     resp = gmail.getProfile(userId="me").execute()
-    slack.chat_postMessage(channel=slack_channel, text=resp["emailAddress"])
-    msg = f"Total no. of messages: `{resp['messagesTotal']}`"
-    slack.chat_postMessage(channel=slack_channel, text=msg)
-    msg = f"Total no. of threads: `{resp['threadsTotal']}`"
-    slack.chat_postMessage(channel=slack_channel, text=msg)
-    msg = f"Current History record ID: `{resp['historyId']}`"
-    slack.chat_postMessage(channel=slack_channel, text=msg)
+    print(resp["emailAddress"])
+    print(f"Total no. of messages: {resp['messagesTotal']}")
```
daabr: In the case of these prints here, and error prints below, I think `print("xxx:", yyy)` is better than `print(f"xxx: {yyy}")`

samples/google/gmail/program.py
pashafateev: What makes the most sense to me is to have the README be like the HTTP sample, but keep the docstring as is. lmkwyt

samples/google/gmail/program.py
pashafateev: Done

samples/google/gmail/program.py
pashafateev: Done

daabr (approved):

PR #83, opened 2024-10-03
chore: remove slack from openai sample

pashafateev:
The sample uses a URL parameter for input, which simplifies the implementation but might not be realistic for more complex prompts. I have two options for addressing this:

- **Option 1**: Document the current approach, informing users of its limitations with URL parameters for input.
- **Option 2**: Rewrite the sample to accept input through the request body, which would introduce slightly more complexity but offer greater flexibility.

Additionally, I’ve adjusted how user input influences the prompt. Previously, with the URL parameter approach, the input wasn’t impacting ChatGPT’s responses as expected. Now, users can directly enter their own prompt and see its effect on the output more clearly. If we switch to using a JSON body, we might be able to revert to the previous method of prompting ChatGPT, as longer inputs could have a more noticeable impact.

### DIFF
--- a/samples/openai_chatgpt/program.py
+++ b/samples/openai_chatgpt/program.py
@@ -1,40 +1,36 @@
 """This program demonstrates AutoKitteh's OpenAI ChatGPT integration.
 
-This program implements a single entry-point function, which is
-configured in the "autokitteh.yaml" manifest file as the receiver
-of Slack "slash_command" events.
+The program implements a single entry-point function, which is
+configured in the "autokitteh.yaml" manifest file to receive HTTP GET requests.
 
-It sends a couple of requests to the ChatGPT API, and sends the responses
-back to the user over Slack, as well as ChatGPT token usage stats.
+It sends a couple of requests to the ChatGPT API, and returns the responses
+to the user, along with ChatGPT token usage stats.
 
 API details:
 - OpenAI developer platform: https://platform.openai.com/
-- Go client API: https://pkg.go.dev/github.com/sashabaranov/go-openai
+- Python client library: https://github.com/openai/openai-python
 
 This program isn't meant to cover all available functions and events.
 It merely showcases various illustrative, annotated, reusable examples.
-
 """
 
-from autokitteh import openai, slack
+from autokitteh.openai import openai_client
 
 MODEL = "gpt-4o-mini"
 
-chatgpt_client = openai.openai_client("chatgpt_conn")
-slack_client = slack.slack_client("slack_conn")
+chatgpt_client = openai_client("chatgpt_conn")
 
 
-def on_slack_slash_command(event):
-    """https://api.slack.com/interactivity/slash-commands
+def on_http_get(event):
+    """Entry-point function for workflow.
 
-    To use the slash command, simply type `/command-name` in the Slack message input,
-    where `command-name` is the name you have assigned to the command in your app.
-    This command does not require any additional text or arguments.
+    example URL: "http://localhost:9980/webhooks/<webhook_slug>?text=autokittens"
 
     Args:
-        event: Slack event data.
+        event: HTTP event data (including URL query parameters).
     """
-
+    params = event.data.url.query
+    text = params.get("text", "")
     # Example 1: trivial interaction with ChatGPT.
     msg = {"role": "user", "content": "Meow!"}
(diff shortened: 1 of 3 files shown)


samples/openai_chatgpt/program.py
```diff
@@ -1,40 +1,36 @@
...
 
-This program implements a single entry-point function, which is
-configured in the "autokitteh.yaml" manifest file as the receiver
-of Slack "slash_command" events.
+The program implements a single entry-point function, which is
+configured in the "autokitteh.yaml" manifest file to receive HTTP GET requests.
 
-It sends a couple of requests to the ChatGPT API, and sends the responses
-back to the user over Slack, as well as ChatGPT token usage stats.
+It sends a couple of requests to the ChatGPT API, and returns the responses
```
daabr: Better be more specific - it prints the responses in the AutoKitteh session log

samples/openai_chatgpt/program.py
```diff
@@ -51,13 +47,11 @@ def on_slack_slash_command(event):
     ]
     msgs = [
         {"role": "system", "content": contents[0]},
-        {"role": "user", "content": contents[1]},
-        {"role": "user", "content": event.data["text"]},
+        {"role": "user", "content": text if text else contents[1]},
```
daabr: 👍

samples/openai_chatgpt/program.py
```diff
@@ -51,13 +47,11 @@ def on_slack_slash_command(event):
...
     ]
 
     resp = chatgpt_client.chat.completions.create(model=MODEL, messages=msgs)
 
-    id = event.data["user_id"]
     for choice in resp.choices:
-        slack_client.chat_postMessage(channel=id, text=choice.message.content)
-    slack_client.chat_postMessage(channel=id, text=f"Usage: `{str(resp.usage)}`")
+        print(choice.message.content)
+    print(f"Usage: `{str(resp.usage)}`")
```
daabr: I think f-strings automatically use `str()` (or `repl()` if you specify `!r`), so no need to call it explicitly... but I could be wrong, which is why it was there to begin with... Please check it out

Update: you can use `{...!s}` instead of `{str(...)}`

samples/openai_chatgpt/program.py
```diff
@@ -1,40 +1,36 @@
...
-    This command does not require any additional text or arguments.
+    example URL: "http://localhost:9980/webhooks/<webhook_slug>?text=autokittens"
 
     Args:
-        event: Slack event data.
+        event: HTTP event data (including URL query parameters).
     """
-
+    params = event.data.url.query
+    text = params.get("text", "")
```
daabr: Usually it's much better to initialize variables as close as possible to where they're being used, unless they have some larger meaning or impact on preceding code. Also, as commented in a different PR, no need for `params`.

daabr (commented):
I like your changes, and using JSON would indeed be a robust solution (i.e. passing prompts and params instead of just a single text string), but supporting HTTP POST with a text body would be easier and actually cover a simple user's needs.

Specifically, limit the trigger to POST and the relevant content type (see the HTTP sample's manifest for examples), and document in the readme and/or code how to send it with curl:

```
curl -X POST "webhook-URL" -H "Content-Type: text/plain" -d "text"
```

samples/openai_chatgpt/program.py
```diff
@@ -1,40 +1,40 @@
...
+    - Curl command: 
+      curl -X POST "<URL>" -H "Content-Type: text/plain" -d "Meow"
 
     Args:
-        event: Slack event data.
+        event: The HTTP event containing request data.
     """
-
+    body = event.data.body.bytes.decode("utf-8")
+    print(body)
```
daabr, line 37:
```suggestion
    print(body)

```

samples/openai_chatgpt/program.py
```diff
@@ -51,13 +51,11 @@ def on_slack_slash_command(event):
     ]
     msgs = [
         {"role": "system", "content": contents[0]},
-        {"role": "user", "content": contents[1]},
-        {"role": "user", "content": event.data["text"]},
+        {"role": "user", "content": body if body else contents[1]},
```
daabr:
```suggestion
        {"role": "user", "content": body or contents[1]},
```

daabr (approved):

PR #84, opened 2024-10-03
chore: remove slack from sheets sample

pashafateev:


samples/google/sheets/autokitteh.yaml
```diff
@@ -9,10 +9,8 @@ project:
...
-    - name: slack_conn
-      integration: slack
   triggers:
-    - name: slack_slash_command
-      connection: slack_conn
-      event_type: slash_command
-      call: program.py:on_slack_slash_command
+    - name: on_http_get
+      type: webhook
+      call: program.py:on_http_get
```
daabr:
```suggestion
    - name: http_get
      type: webhook
      event_type: get
      call: program.py:on_http_get
```

samples/google/sheets/program.py
```diff
@@ -5,52 +5,39 @@
...
 sheet = google_sheets_client("sheets_conn").spreadsheets().values()
-slack = slack_client("slack_conn")
 
 
-def on_slack_slash_command(event):
+def on_http_get(event):
     """Use a Slack slash command to interact with a Google Sheet.
 
     See: https://api.slack.com/interactivity/slash-commands, and
     https://api.slack.com/interactivity/handling#message_responses
```
daabr: Need to update this docstring too to refer to HTTP instead of Slack. Also, it's good that you simplified the documentation and code by getting rid of the full URL option.

samples/google/sheets/program.py
```diff
@@ -5,52 +5,39 @@
...
-    - A full Google Sheets URL (to extract the spreadsheet ID from it)
+    - A valid Google Sheets ID (https://developers.google.com/sheets/api/guides/concepts)
 
     Args:
         event: Slack event data.
     """
-    user_id = event.data.user_id
-    sheet_id = _extract_sheet_id(event.data.text, user_id)
+    params = event.data.url.query
+    sheet_id = params.get("sheet_id")
```
daabr: No need to clarify what `event.data.url.query` is, it's clear (and short) enough.

Also, while `sheet_id` is descriptive, there's no need to be that specific in the URL when we're accepting only one type of ID.

```suggestion
    sheet_id = event.data.url.query.get("id")
```

samples/google/sheets/program.py
```diff
@@ -72,11 +59,11 @@ def _write_values(spreadsheet_id, slack_target):
 
     text = f"Updated: range `{resp['updatedRange']!r}`, `{resp['updatedRows']}` rows, "
     text += f"`{resp['updatedColumns']}` columns, `{resp['updatedCells']}` cells"
-    slack.chat_postMessage(channel=slack_target, text=text)
+    print(text)
```
daabr: Since we're using `print` instead Slack, it's more readable to use multiple prints instead of a single long formatted string.

For example:

```
print(f"Updated range: {resp['updatedRange']!r}")
print("Rows:", resp["updatedRows"])
print("Columns:", resp["updatedColumns"])
print("Cells:", resp["updatedCells"])

File: samples/google/sheets/program.py [hunk 5]
```diff
@@ -94,11 +81,11 @@ def _read_values(id, slack_target):
         data_type, formatted, unformatted = row
         text = "Row {0}: {1} = formatted `{2!r}`, unformatted `{3!r}`"
         text = text.format(i + 1, data_type, formatted, unformatted)
-        slack.chat_postMessage(channel=slack_target, text=text)
+        print(text)
```
daabr, line 83: No need for the `text` helper var anymore, a single `print(f"...") is short enough, I think, and therefore better

Btw, IMO the `value` helper var in the functions below is still useful, in contrast (because setting is a little complicated)

samples/google/sheets/program.py
pashafateev: Oops! Done

samples/google/sheets/program.py
pashafateev: Done

samples/google/sheets/program.py
pashafateev: python forces to two lines, so I kept the `text` var

samples/google/sheets/program.py
pashafateev: Done

samples/google/sheets/program.py
```diff
@@ -5,52 +5,37 @@
...
 
-def on_slack_slash_command(event):
-    """Use a Slack slash command to interact with a Google Sheet.
+def on_http_get(event):
+    """Entry point for the workflow.
 
-    See: https://api.slack.com/interactivity/slash-commands, and
-    https://api.slack.com/interactivity/handling#message_responses
+    This function expects the URL parameter 'id' to be:
+    - A valid Google Sheets ID (see https://developers.google.com/sheets/api/guides/concepts)
```
daabr:
```suggestion
    This function expects the URL parameter 'id' to be a valid Google Sheets ID
    (see https://developers.google.com/sheets/api/guides/concepts).
```

samples/google/sheets/program.py
```diff
@@ -5,52 +5,37 @@
...
 
     Args:
-        event: Slack event data.
+        event: HTTP event data, including URL query parameters.
     """
-    user_id = event.data.user_id
-    sheet_id = _extract_sheet_id(event.data.text, user_id)
+    sheet_id = event.data.url.query.get("id")
     if not sheet_id:
         return
```
daabr, line 28: Maybe also print a message that a sheet ID is required but not found?

daabr (approved):
Approved with 2 tiny cosmetic nits

samples/google/sheets/program.py
pashafateev: Done

PR #87, opened 2024-10-06
feat: add jira ticketing handling workflow

pashafateev:
**Known Limitations:**

- **Slack Invitations:** Due to a Jira permissions issue that requires further investigation, only the creator of the issue is automatically invited to the Slack channel. All other participants must be added manually.
- **Channel Filtering:** Adding a channel filter alongside the mention filter—specifically using `'app_mention' && data.event.channel == 'C07QKNEF610'`—did not work. As a workaround, the event is checked within the Python code.  
  ^ The issue is we haven't decided on syntax which is being tracked [here](https://linear.app/autokitteh/issue/ENG-1654/bug-unable-to-add-filter-for-a-slack-event-using-subscribenext-event).  
  I'm open to changing it to the current filter syntax, but that means it would potentially break once that change merges.

### DIFF
--- a/jira_ticket_handling/README.md
+++ b/jira_ticket_handling/README.md
@@ -1,0 +1,89 @@
+
+# Jira-Slack Issue Workflow Automation
+
+This project automates the process of creating Slack channels for new Jira issues and streamlines issue management. It listens for new Jira issues, creates a dedicated Slack channel for each issue, invites the issue creator to the channel, and waits for the user to confirm completion by mentioning the Slack app with `@ak done`. Once confirmed, it archives the channel, notifies the channel members, and updates the Jira issue status to `DONE`.
+
+> [!IMPORTANT]
+> The command `@ak done` assumes that you have named your custom Slack app "ak." If you used a different name, be sure to substitute "ak" with the appropriate name of your Slack app when mentioning it in a Slack channel.
+
+## How It Works
+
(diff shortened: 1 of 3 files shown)


jira_ticket_handling/README.md
```diff
@@ -0,0 +1,89 @@
+
```
daabr:
```suggestion
```

jira_ticket_handling/README.md
```diff
@@ -0,0 +1,89 @@
...
+# Jira-Slack Issue Workflow Automation
+
+This project automates the process of creating Slack channels for new Jira issues and streamlines issue management. It listens for new Jira issues, creates a dedicated Slack channel for each issue, invites the issue creator to the channel, and waits for the user to confirm completion by mentioning the Slack app with `@ak done`. Once confirmed, it archives the channel, notifies the channel members, and updates the Jira issue status to `DONE`.
+
+> [!IMPORTANT]
+> The command `@ak done` assumes that you have named your custom Slack app "ak." If you used a different name, be sure to substitute "ak" with the appropriate name of your Slack app when mentioning it in a Slack channel.
```
daabr: Let's use `@autokitteh` as a default instead of `@ak`.

pashafateev: Done

PR #88, opened 2024-10-07
feat: port reviewkitteh to python

pashafateev:


reviewkitteh/program.py
```diff
@@ -0,0 +1,63 @@
...
+            log(f"Meowing at {the_chosen_one}")
+
+            user_email = f"{the_chosen_one}@{ORG_DOMAIN}"
+            user = slack.users_lookupByEmail(email=user_email)["user"]
+            msg = f"Paging <@{user["id"]}>"
+            slack.chat_postMessage(channel=CHANNEL_ID, text=msg, thread_ts=ts)
+
+
+def log(msg):
+    print(f"[{time.strftime('%Y-%m-%d %H:%M:%S')}] {msg}")
```
daabr: Simpler (and with microseconds):
```
from datetime import datetime

print(f"[{datetime.now()}] {msg}")
```

reviewkitteh/program.py
```diff
@@ -0,0 +1,63 @@
+"""A program that listens for GitHub pull requests and meows at random people.
```
daabr:
```suggestion
"""Listen for GitHub pull requests and meow at random people.
```

reviewkitteh/program.py
```diff
@@ -0,0 +1,63 @@
...
+
+github = github_client("github_conn")
+googlesheets = google_sheets_client("googlesheets_conn").spreadsheets().values()
+slack = slack_client("slack_conn")
+
+
+def on_github_pull_request(event):
+    """Workflow's entry-point."""
+    if event.data.action not in {"opened", "reopened"}:
+        return
```
daabr: Why not move this to the a filter in the manifest?

reviewkitteh/README.md
```diff
@@ -16,7 +16,7 @@ This project automates the process of listening for GitHub pull request events a
 ### Configure integrations
 
 > [!IMPORTANT]
->  The `autokitteh.yaml` file includes environment variables for the GitHub, Slack, and Google Sheets connections that need to be configured.
+>  The `autokitteh.yaml` file includes environment variables for the GitHub, Slack, and Google Sheets connections that need to be configured. When setting the `CHANNEL` variable, be sure to use the channel ID, not the channel name.
```
daabr, line 19: 1. The manifest has `CHANNEL_ID`, not `CHANNEL` like in the original project
2. But why just an ID? Are you sure a name wouldn't work?

pashafateev: Yes, apparently `chat.update` only accepts ID's although this isn't well documented. chatGPT is the one who let me know about this. I couldn't find much online, but I did find [this](https://github.com/slack-ruby/slack-ruby-client/issues/270#issuecomment-492279873).

The [official Slack documentation](https://api.slack.com/methods/chat.update) doesn't mention anything, although it does use an ID and not a name (which isn't saying much considering name works for `chat.postMessage`).

All I know, is `chat.update` couldn't find the channel until I fed it an ID.

reviewkitteh/program.py
pashafateev: Done

reviewkitteh/program.py
pashafateev: Good idea. Done

reviewkitteh/program.py
```diff
@@ -0,0 +1,62 @@
...
+slack = slack_client("slack_conn")
+
+
+def on_github_pull_request(event):
+    """Workflow's entry-point."""
+    pr = event.data.pull_request
+    msg = f"{pr.html_url} [{pr.state}]"
+    ts = slack.chat_postMessage(channel=CHANNEL_ID, text=msg)["ts"]
+
+    for i in itertools.count(start=1):
```
daabr, line 36: It's not strictly wrong to use a `for` loop with `if`, but the original code is simpler (`i=0` + `while` loop)

reviewkitteh/program.py
```diff
@@ -0,0 +1,62 @@
...
+        if pr.state in {"closed", "merged"}:
```
daabr, line 37: tl;dr - use a tuple (`()`) instead of a set (`{}`)

---

Performance optimization deep-dive:

When using `if x in y`, it is smart to use a set (`{1, 2}`) like you do instead of a list (`[1, 2]`) like in the original code, but that performance advantage is relevant only for a **large** number of elements and a **large** number of checks, where the O(1) search time of sets becomes relevant compared to O(n) in lists.

For tiny sets like here, the actual search time is effectively the same, but sets use more memory than lists. So lists are better than sets here.

Even better (again, in this specific case) - a tuple: same runtime complexity for operations, but uses less memory overall than a list, because it's immutable. And in our case this collection is indeed immutable.

---

Also note that re-initializing the exact same set/list/tuple in every iteration of the loop is unnecessary and wasteful. You could initialize it once before the loop, or as a module-level variable, but...

Compare `if x in var_from_somewhere_else` to `if x in (1, 2)`: the second version is less optimal in a loop, but is considered more readable. In other words, this is a common example where we intentionally prefer less optimal but more readable code, because the value of readability in this case outweighs the speed/space gains.

daabr (approved):

reviewkitteh/program.py
pashafateev: I agree with that. My reasoning behind using the `itertools.count` is apparently it's more pythonic, at least according to the research I did. I'll change it back, but my thought was if we're targeting python users, this might be more familiar? On the other hand if we're targeting very basic python users, then `while` might be a friendlier option?

reviewkitteh/program.py
pashafateev: Done. Thanks for the deep dive. Very helpful.

PR #104, opened 2024-10-25
feat: port github_copilot_seats to python

pashafateev:
`_add_seat`, `_remove_seat`, and `_get_all_seats()` utilize the `Requester` object to make requests to APIs that PyGithub doesn't currently support. To facilitate this, I updated the `pyproject.toml` to `"PyGithub @ git+https://github.com/PyGithub/PyGithub.git"` in a separate [PR](https://github.com/autokitteh/autokitteh/pull/802), ensuring we get the latest version. This change is necessary because the `Requester` object was recently added, and the last release was in August.

I mention this here because both PRs will need to be merged simultaneously.


github_copilot_seats/triggers.py
```diff
@@ -0,0 +1,32 @@
+from autokitteh.slack import slack_client
+from seats import prune_idle_seats, find_idle_seats
+
+s = slack_client("slack_conn")
+
+
+def on_schedule() -> None:
+    "Entry point for the workflow"
```
itayd: i think it's customary to have `"""` for docstrings. https://peps.python.org/pep-0257/#one-line-docstrings

itayd: i would also either remove the docstring or put something more descriptive.

github_copilot_seats/triggers.py
```diff
@@ -0,0 +1,32 @@
...
+    print(prune_idle_seats())
+
+
+def on_slack_slash_command(event) -> None:
+    "Entry point for the workflow"
```
itayd: ditto

github_copilot_seats/triggers.py
```diff
@@ -0,0 +1,32 @@
...
+    cmd = event.data.text
+    cid = event.data.channel_id
+
+    if cmd not in {"prune-idle-copilot-seats", "find-idle-copilot-seats"}:
```
itayd: never seen `{}` used in this way, it think it means a set? no need to be fancy here. i think `[]` or `{}` is more customary.

github_copilot_seats/triggers.py
```diff
@@ -0,0 +1,32 @@
...
+        s.chat_postMessage(channel=cid, text="Unrecognized command")
```
itayd: i wonder if we're using multiple automations on the same bot, this would give you multiple "unrecognized command" if implemented this way in multiple workflows. @daabr wdyt?

github_copilot_seats/triggers.py
```diff
@@ -0,0 +1,32 @@
...
+        return
+
+    if cmd == "prune-idle-copilot-seats":
+        seats = prune_idle_seats()
+        msg = f"Engaged {len(seats)} new idle seats: {', '.join(get_logins(seats))}"
```
itayd, line 23:
```suggestion
        msg = f"Engaged {len(seats)}. new idle seats: {', '.join(get_logins(seats))}"
```

github_copilot_seats/triggers.py
```diff
@@ -0,0 +1,32 @@
...
+        s.chat_postMessage(channel=cid, text=msg)
+    elif cmd == "find-idle-copilot-seats":
+        seats = find_idle_seats()
+        msg = f"Found {len(seats)} idle seats: {', '.join(get_logins(seats))}"
+        s.chat_postMessage(channel=cid, text=msg)
+
```
itayd, line 27: instead of doing the fancy set above, if we decide to keep it I would move it as an `else` here.

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,122 @@
...
+
+logins = LOGINS.split(",")
+github = github_client("github_conn")
+slack = slack_client("slack_conn")
+
+
+def prune_idle_seats() -> list:
+    """Prunes idle GitHub Copilot users based on their last activity time."""
+    seats = find_idle_seats()
+    new_idle_seats = []
```
itayd: why is this needed?

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,122 @@
...
+    for seat in seats:
+        new_idle_seats.append(seat)
+        autokitteh.start(loc="seats.py:engage_seat", data={"event": {"seat": seat}})
```
itayd: do we need the `"event": ...` here? not sure.

github_copilot_seats/seats.py
itayd: just use `seats`.

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,122 @@
...
+IDLE_USAGE_THRESHOLD = int(os.getenv("IDLE_USAGE_THRESHOLD"))
+LOGINS = os.getenv("LOGINS")
+LOG_CHANNEL = os.getenv("LOG_CHANNEL")
+
+logins = LOGINS.split(",")
+github = github_client("github_conn")
+slack = slack_client("slack_conn")
+
+
+def prune_idle_seats() -> list:
```
itayd: can you specify the type of list in `list[<TYPE_OF_SEAT>]`?

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,122 @@
...
+    """Prunes idle GitHub Copilot users based on their last activity time."""
+    seats = find_idle_seats()
+    new_idle_seats = []
+    for seat in seats:
+        new_idle_seats.append(seat)
+        autokitteh.start(loc="seats.py:engage_seat", data={"event": {"seat": seat}})
+    return new_idle_seats
+
+
+def find_idle_seats() -> list:
```
itayd: ditto re type

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,122 @@
...
+            f"{'>=' if is_idle else '<'} {IDLE_USAGE_THRESHOLD}"
+        )
+
+        if is_idle:
+            idle_seats.append(seat)
+
+    return idle_seats
+
+
+def engage_seat(seat: dict) -> None:
```
itayd: please specify the full dict type `dict[K, V]`

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def _get_all_seats() -> list:
+    # TODO: pagination.
+    url = f"https://api.github.com/orgs/{GITHUB_ORG}/copilot/billing/seats"
+    _, data = github.requester.requestJsonAndCheck("GET", url)
+    return data["seats"]
+
+
+def log(github_login: str, msg: str) -> None:
```
itayd: `log` is a bit confusing here. maybe `report`?

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,122 @@
...
+    Args:
+        seat (dict): Contains details about the assigned GitHub user.
+    """
+    github_login = seat["assignee"]["login"]
+
+    log(github_login, "engaging")
+
+    slack_id = github_username_to_slack_user_id(github_login, GITHUB_ORG)
+
+    if not slack_id:
```
itayd:
```suggestion
    if not slack_id:
```

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+
+        if is_idle:
+            idle_seats.append(seat)
+
+    return idle_seats
+
+
+def engage_seat(seat: dict) -> None:
+    """Engages a GitHub user assigned to a seat by identifying their corresponding Slack user and initiating a workflow.
```
itayd, line 64: specify in the log that this is meant to be run as a child workflow/session. (right?)

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # Based on the user's response, it either confirms the action or reinstates the seat
+    if value == "ok":
+        slack.chat_postMessage(slack_id, "Okey dokey!")
+        log(github_login, "ok")
+    elif value == "reinstate":
+        log(github_login, "reinstate")
+        _add_seat(github_login)
+        msg = "You have been reinstated to the Copilot program."
+        slack.chat_postMessage(slack_id, msg)
```
itayd:
```suggestion
        slack.chat_postMessage(slack_id, "You have been reinstated to the Copilot program.")
```

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,122 @@
...
+    else:
+        log(github_login, f"weird response: {value}")
```
itayd: don't you want to notify the person that their response is invalid?

github_copilot_seats/seats.py
pashafateev: Done

github_copilot_seats/seats.py
pashafateev: Done

github_copilot_seats/seats.py
pashafateev: That's what you told me last time we talked about it. When I do data={"seat": seat} nothing changes, but I don't think I've ever entered that part of the code because I don't want to actually kick somebody out.

github_copilot_seats/seats.py
pashafateev: Done

github_copilot_seats/seats.py
pashafateev: Done

github_copilot_seats/seats.py
pashafateev: Done

github_copilot_seats/seats.py
pashafateev: It will autoformat to this
```        
slack.chat_postMessage(
    slack_id, msg="You have been reinstated to the Copilot program."
)
```

github_copilot_seats/seats.py
pashafateev: have both direct message and report to channel?

github_copilot_seats/seats.py
pashafateev: Done

github_copilot_seats/triggers.py
pashafateev: Done

github_copilot_seats/triggers.py
pashafateev: Done

github_copilot_seats/triggers.py
pashafateev: Done

github_copilot_seats/triggers.py
pashafateev: I think you're right about that.

github_copilot_seats/triggers.py
pashafateev: Done

github_copilot_seats/README.md
```diff
@@ -11,10 +11,18 @@ This automation searches daily for all users in a GitHub organization that are a
 If Copilot was not used in a preceding period, it automatically unregisters them, and then notifies them.
 Users can then optionally ask for their subscription to be reinstated.
 
+## Before Deploying This AutoKitteh Project:
```
daabr, line 14: Keep 1 empty line above and below headers.
```suggestion
## Before Deploying This AutoKitteh Project:

```

github_copilot_seats/README.md
```diff
@@ -11,10 +11,18 @@ This automation searches daily for all users in a GitHub organization that are a
...
+- Set the `GITHUB_ORG` in the project's vars.
```
daabr, line 16: I think we can determine the org automatically (where the GitHub app was installed) - but that's for a later PR.

github_copilot_seats/triggers.py
```diff
@@ -0,0 +1,28 @@
+from autokitteh.slack import slack_client
+from seats import prune_idle_seats, find_idle_seats
```
daabr: Prefer to import modules instead of classes/functions within them, unless the module name is cumbersome/redundant.

So a nice way to improve your code would be to use `import seats`, and rename the functions in it, so calls here would be `seats.prune_idle` and `seats.find_idle`.

github_copilot_seats/triggers.py
```diff
@@ -0,0 +1,28 @@
...
+
+s = slack_client("slack_conn")
+
+
+def on_schedule() -> None:
```
daabr, line 8: Just making sure: schedular-based handler functions shouldn't have a parameter, like `event` in connection-based handlers?

github_copilot_seats/triggers.py
```diff
@@ -0,0 +1,28 @@
...
+    print(prune_idle_seats())
+
+
+def on_slack_slash_command(event) -> None:
+    cmd = event.data.text
```
daabr: Cheap normalization for extra safety:
```suggestion
    cmd = event.data.text.lower()
```

github_copilot_seats/triggers.py
```diff
@@ -0,0 +1,28 @@
...
+    if cmd == "prune-idle-copilot-seats":
+        seats = prune_idle_seats()
+        msg = f"Engaged {len(seats)} new idle seats: {', '.join(get_logins(seats))}"
+        s.chat_postMessage(channel=cid, text=msg)
+    elif cmd == "find-idle-copilot-seats":
+        seats = find_idle_seats()
+        msg = f"Found {len(seats)} idle seats: {', '.join(get_logins(seats))}"
+        s.chat_postMessage(channel=cid, text=msg)
+    else:
+        s.chat_postMessage(channel=cid, text="Unrecognized command")
```
daabr: `chat_postMessage` isn't necessarily bad, but note that slash commands are global (not limited to specific channels), and not visible to other users in that channel.

This means that there are 2 better options to respond to slash commands, choose whichever you prefer:

1. Use the user ID as the channel ID, so the app will send a (personal) DM to the user, regardless of the channel where they typed the command
2. Send an ephemeral message to the channel, instead of a regular message, so only the user sees the response

github_copilot_seats/triggers.py
daabr: Yes. Ideally, people will use custom OAuth after their initial experiments.

github_copilot_seats/seats.py
daabr: Either way is fine. Just know that it's a legitimate response to Itay's comment that you want to avoid multi-line wrapping due to auto-formatting. Up to you.

github_copilot_seats/helpers.py
```diff
@@ -0,0 +1,53 @@
...
+slack = slack_client("slack_conn")
+
+
+def _email_to_slack_user_id(email: str) -> str:
+    """Fetch Slack user ID based on email address."""
+    resp = slack.users_lookupByEmail(email=email)
+    return resp["user"]["id"] if resp.get("ok", False) else None
+
+
+def github_username_to_slack_user_id(username: str, owner_org: str) -> str:
```
daabr: Unused arg
```suggestion
def github_username_to_slack_user_id(username: str) -> str:
```

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,113 @@
+import json
+import os
+from datetime import datetime, timezone, timedelta
+from typing import Any
+import autokitteh
+from autokitteh.github import github_client
+from autokitteh.slack import slack_client
+from helpers import github_username_to_slack_user_id
+
```
daabr, line 11: Black/Ruff won't complain or fix this automatically, but let's be consistent about 2 empty lines between imports and everything else, just line Black/Ruff **do** automatically enforce 2 empty lines between module-level functions.
```suggestion


```

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,113 @@
...
+GITHUB_ORG = os.getenv("GITHUB_ORG")
```
daabr: If I'm not mistaken, all references to a GitHub org are now obsolete. We needed them only in the Starlark wrapper for GitHub, to distinguish authenticated API calls from anonymous ones.

github_copilot_seats/README.md
pashafateev: Do you want me to open a ticket?

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,113 @@
...
+IDLE_USAGE_THRESHOLD = int(os.getenv("IDLE_USAGE_THRESHOLD"))
+LOGINS = os.getenv("LOGINS")
+LOG_CHANNEL = os.getenv("LOG_CHANNEL")
+
+logins = LOGINS.split(",") if LOGINS else None
```
daabr: 1. Why not put this condition directly in the assignment to `LOGINS` in line 12?
2. Why `else None`? If you remove it, and the env var is empty, you get an empty list - which doesn't cause loops to crash, and also evaluates to `False` if checked as a Boolean

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,113 @@
...
+            print(f"skipping {seat.assignee.login}")
+            continue
+
+        delta = t - seat.last_activity_at
+        is_idle = delta >= timedelta(minutes=IDLE_USAGE_THRESHOLD)
+
+        print(
+            f"{seat.assignee.login}: {t} - {seat.last_activity_at} = {delta} "
+            f"{'>=' if is_idle else '<'} {IDLE_USAGE_THRESHOLD} minutes"
+        )
```
daabr: It's more readable in this case to use helper vars to avoid multi-line wrapping of a single `print` statement

github_copilot_seats/triggers.py
pashafateev: I'm fine with using modules instead, but I don't like the renaming of the functions. Yes, it adds brevity to `triggers.py` but it sacrifices clarity in `seats.py`.

I'd rather do `seats.find_idle_seats` etc

github_copilot_seats/seats.py
pashafateev: you're saying I can get the org from the python client? I need the org name to get the copilot object

github_copilot_seats/triggers.py
pashafateev: I will check

github_copilot_seats/README.md
daabr: No - see comment below about probably not needing orgs at all.

daabr: Whichever you prefer - either just add it now, or open a ticket as a reminder to do this sometime soon. Either way it should be simple.

github_copilot_seats/seats.py
daabr: Oops, I missed `github.get_organization`. Anyway, yes, I think so. But feel free to check this out at a later time.

github_copilot_seats/seats.py
pashafateev: Do you mean like this?
```
LOGINS = os.getenv("LOGINS").split(",")
```
That doesn't work, because instead of an empty list, I get `['']`

github_copilot_seats/triggers.py
pashafateev: importing seats created a name conflict. take a look at the solution, I think what we had before was fine considering it's a small project, but I stand by not changing the function names in `seats.py`.

github_copilot_seats/triggers.py
pashafateev:
```
[plan] trigger "github_copilot_seats:connection:github_copilot_seats/slack_conn/check_now": not as desired, will update
Error: action update_trigger github_copilot_seats:schedule:@daily/check_daily: update schedule: schedule: update scheduler workflow: workflow not found for ID: temporal-sys-scheduler:trg_01jg5aj8mqfnetdc0wtjqjnh83
```

This is the error you get if you try to add an event parameter.

github_copilot_seats/triggers.py
pashafateev: Done

github_copilot_seats/triggers.py
daabr: I still prefer "import seats" instead of importing the functions directly, and I think your solution (renaming the "seats" variables) is fine. If it was more complicated/convoluted than that then that would indeed be a good reason to keep the original imports.

Either way, I agree with your rationale to keep "seats" in the function names in the `seats` module.

github_copilot_seats/triggers.py
daabr: Not sure I understand this error message - I was asking if the Python function that handles the scheduler trigger needs to have an `event` parameter like regular trigger entry points, or if it's fine without it like in your code.

But I'm assuming you tested triggering these events now, and if that works it answers my question.

github_copilot_seats/triggers.py
```diff
@@ -0,0 +1,28 @@
+from autokitteh.slack import slack_client
+import seats
+
+s = slack_client("slack_conn")
```
daabr: Order the imports, and split into 3 blocks: standard library (none in this specific file), 3rd-party (including AK SDK), and project-internal.

```suggestion
from autokitteh.slack import slack_client

import seats

s = slack_client("slack_conn")
```

github_copilot_seats/seats.py
daabr: You're right! In that case, add `else []`, but not `else None` - so at least `logins` will always have the same type.

It doesn't actually help to simplify line 39, but it's still safer.

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,113 @@
+import json
+import os
+from datetime import datetime, timezone, timedelta
+from typing import Any
+import autokitteh
+from autokitteh.github import github_client
+from autokitteh.slack import slack_client
+from helpers import github_username_to_slack_user_id
```
daabr: Order the imports, and split into 3 blocks: standard library, 3rd-party (including AK SDK), and project-internal.

```suggestion
from datetime import datetime, timezone, timedelta
import json
import os
from typing import Any

import autokitteh
from autokitteh.github import github_client
from autokitteh.slack import slack_client

from helpers import github_username_to_slack_user_id
```

github_copilot_seats/seats.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+GITHUB_ORG = os.getenv("GITHUB_ORG")
```
daabr: How to get the GitHub org name without a manually-specified project var - it's already a connection var:

https://github.com/autokitteh/kittehub/pull/139/files#diff-81e3c3caea8e0d76885489976911b513696a4ee6b1084e874f4f6ae0a4c37216

daabr (approved):
Approved with a couple of tiny nits, but note the lint error (need for format `seats.py`).

If the string is still too long, you can also split it into 2 parts, e.g. https://github.com/autokitteh/kittehub/blob/main/purrr/slack_channel.py#L105-L106 - ultimately whichever formatting you prefer is fine, that's a personal stylistic choice, as long as it's not hard to understand.

PR #105, opened 2024-10-25
fix: change env var name to match yaml

pashafateev:
changed var name in `.yaml` in a code review and forgot to change in `program.py`

### DIFF
--- a/google_cal_to_asana/autokitteh.yaml
+++ b/google_cal_to_asana/autokitteh.yaml
@@ -7,7 +7,7 @@ project:
   name: google_cal_to_asana
   vars:
     - name: ASANA_PROJECT_GID
-      value: 1208402622198831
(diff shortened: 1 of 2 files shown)


google_cal_to_asana/autokitteh.yaml
```diff
@@ -7,7 +7,7 @@ project:
   name: google_cal_to_asana
   vars:
     - name: ASANA_PROJECT_GID
-      value:
+      value: 1208402622198831
```
itayd, line 10: remove

itayd (approved):

google_cal_to_asana/autokitteh.yaml
itayd:
```suggestion
      value:
```

PR #108, opened 2024-10-29
feat(pipeline): A option to create database from workflow

tebeka:


data_pipeline/pipeline.py
```diff
@@ -1,17 +1,18 @@
...
+from contextlib import closing
+from io import BytesIO
+from pathlib import Path
 
 import autokitteh
 from autokitteh.aws import boto3_client
 
-
-DB_DSN = os.getenv("DB_DSN")  # Secret
+DB_DSN = os.getenv('DB_DSN')  # Secret
```
pashafateev: **Curiosity:** This connects the user to our instance of sqlite? And if running locally they can set this to something else to connect to their instance?

data_pipeline/pipeline.py
```diff
@@ -1,17 +1,18 @@
...
+CREATE_DB = os.getenv('CREATE_DB', 'no').lower() in {'y', 'yes', 'true'}
```
pashafateev: How is `CREATE_DB` set? I don't see it in the `autokitteh.yaml`? The user could add a new variable themselves, but I think it's more clear if it comes preconfigured.

data_pipeline/pipeline.py
tebeka: SQLite is an embedded database, basically a file. If they want to connect to a specific instance they need to make it accessible to the workflow (currently not possible).
This is just an example, real life pipelines will connect to PostgreSQL, ClickHouse or other client/server database.

data_pipeline/pipeline.py
tebeka: Good idea, can you do it? (not working until Sunday)

pashafateev: Sure, but I have another question. If user is on cloud and sets `CREATE_DB` = 'yes', then what is the `DB_DSN`? I don't think it's clear in this workflow what the user should do with that env.

I think we either need to add some documentation specifying what file paths are available, or checking that if `CREATE_DB` is set to 'yes' then defaulting to something like this:
```
   code_dir = Path(__file__).absolute().parent
   DB_DSN = code_dir / "data" / "hikes.db"
```

tebeka: Maybe something like:

```python
code_dir = Path(__file__).parent
DB_DSN = os.getenv('DB_DSN') or code_dir / 'hikes.db'  # Secret
```

I don't want the extra `data` directory since we'll need to create it and it's an extra step that's not relevant to what we're trying to show.

daabr (approved):

PR #116, opened 2024-11-17
ENG-1821: Fixed Twilio sample error 

mario99logic:
fixed ENG-1821 , added a helper function in case the user doesnt add the "+" as URL-encoded

### DIFF
--- a/samples/twilio/program.py
+++ b/samples/twilio/program.py
@@ -37,7 +37,11 @@ def on_http_get(event):
         event (object): An event object containing the request data.
     """
     to = event.data["url"]["query"]["to"]
-    to = add_plus_prefix(to)
+
+    # Add a '+' if missing
+    if not to.startswith("+"):
+        to = f"+{to}"
+
     # Send SMS text via Twilio
     message = t.messages.create(
         from_=FROM_PHONE_NUMBER,
@@ -53,10 +57,3 @@ def on_http_get(event):
         body="This is an AutoKitteh demo message, meow!",
     )
     print(f"WhatsApp message sent: {whatsapp_message.sid}")
-
-
-def add_plus_prefix(phone_number):
-    # Add a '+' if missing
-    if not phone_number.startswith("+"):
-        return f"+{phone_number}"
-    return phone_number
--- a/samples/twilio/program.py
+++ b/samples/twilio/program.py
@@ -33,17 +33,15 @@ def on_http_get(event):
     parameter 'to'. The message will be sent both as an SMS and a WhatsApp
     message to the specified number.
 
-    Note: No need to include the `+` prefix in the `to` parameter.
-    The program automatically appends the `+` where required.
-
     Args:
         event (object): An event object containing the request data.
     """
-    to = event.data["url"]["query"]["to"].strip()
+    to = event.data["url"]["query"]["to"]
+    to = add_plus_prefix(to)
     # Send SMS text via Twilio
     message = t.messages.create(
         from_=FROM_PHONE_NUMBER,
-        to="+" + to,
+        to=to,
         body="This is an AutoKitteh demo message, meow!",
     )
     print(f"SMS message sent: {message.sid}")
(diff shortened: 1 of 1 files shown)


samples/twilio/program.py
```diff
@@ -33,22 +33,25 @@ def on_http_get(event):
...
     message to the specified number.
 
+    Note: No need to include the `+` prefix in the `to` parameter.
+    The program automatically appends the `+` where required.
+
     Args:
         event (object): An event object containing the request data.
     """
-    to = event.data["url"]["query"]["to"]
+    to = event.data["url"]["query"]["to"].strip()
```
daabr: This is unnecessary: it's very rare for people to put spaces (or `%20`) in URLs without a specific reason

samples/twilio/program.py
```diff
@@ -33,22 +33,25 @@ def on_http_get(event):
     parameter 'to'. The message will be sent both as an SMS and a WhatsApp
     message to the specified number.
 
+    Note: No need to include the `+` prefix in the `to` parameter.
+    The program automatically appends the `+` where required.
```
daabr: A question, not a conclusion: *is it* better to not specify the "+" prefix in the number? This is a standard practice for people when they write international phone numbers (obvious example: this is what Twilio's API expects).

samples/twilio/program.py
```diff
@@ -33,22 +33,25 @@ def on_http_get(event):
...
     Args:
         event (object): An event object containing the request data.
     """
-    to = event.data["url"]["query"]["to"]
+    to = event.data["url"]["query"]["to"].strip()
     # Send SMS text via Twilio
     message = t.messages.create(
         from_=FROM_PHONE_NUMBER,
-        to=to,
+        to="+" + to,
```
daabr: Aren't you substituting one problem (the user forgets to write "+") with a different problem? (the user writes "+" as expected, and we mess it up by adding another "+")

Even if we prefer to add "+" automatically (which isn't clear to me yet), we need to accept inputs that already have it, and not change them. Maybe with a helper function that adds a "+" prefix only if it's missing.

samples/twilio/program.py
```diff
@@ -33,22 +33,25 @@ def on_http_get(event):
...
         body="This is an AutoKitteh demo message, meow!",
     )
     print(f"SMS message sent: {message.sid}")
 
     # Send a WhatsApp message to the same number
     whatsapp_message = t.messages.create(
         from_="whatsapp:" + FROM_PHONE_NUMBER,
-        to="whatsapp:" + to,
+        to="whatsapp+:" + to,
```
daabr: Same comment as above, but anyway - is this a typo? Shouldn't it be "whatsapp:+"?

samples/twilio/program.py
pashafateev: what we found is that specifying it without url encoding results in the '+' being converted to ' '. So either way we have to add the '+' ourselves. Or we ask the user to use URL encoding which is kind of clunky imo.

I wasn't happy with either solution honestly, but I thought this was the lesser evil.

samples/twilio/program.py
pashafateev: @daabr see [comment](https://github.com/autokitteh/kittehub/pull/116#discussion_r1847071982)

in case they do put a '+' it will turn into a ' '.

samples/twilio/program.py
pashafateev: See previous comments, but yes, I agree with this IF it's possible for the '+' to pass through without being URL encoded. I am assuming it can, because I tested this when I wrote it and iirc it worked with the '+' included.

mario99logic: The thing is that even if the user adds "+" it won't matter because when we get it from the event we will get only the numbers. because of that reason we decided to add the "+", so even if the user types + or not this code will work

samples/twilio/program.py
daabr:
```suggestion
    to = event.data["url"]["query"]["to"]

    # Fix unescaped/missing "+" prefix.
    to = to if to.startswith("+") else "+" + to.strip()

```

samples/twilio/program.py
daabr: @pashafateev ack on the motivation

@mario99logic it depends - when the "+" is URL-encoded (manually by the user, or automatically by curl/wget/etc.) then it will be received correctly - so see my alternative suggestion above

samples/twilio/program.py
daabr: ...And if we use my suggestion above, the "+" is not needed here either

samples/twilio/program.py
daabr: SGTM. See my recent code suggestion.

daabr (approved):
Discussed offline: bring back the 1-line version of the helper function. Approving in advance.

PR #122, opened 2024-11-27
feat: add script to automate README updates for DOC-16

mario99logic:
This PR introduces a script to automate the process of updating the README.md file. The script ensures that tables in the README are refreshed with new metadata extracted from new samples.

Added a Python script to scan and process README.md files.
•  Implemented features to clean existing tables and replace them with updated content.
•  Linked the script functionality to Linear ticket DOC-16 for tracking.


scripts/create_project_table.py
```diff
@@ -0,0 +1,137 @@
+import os
```
daabr: Please add a module docstring (see the example here, don't add a license): https://google.github.io/styleguide/pyguide.html#382-modules

scripts/create_project_table.py
```diff
@@ -0,0 +1,137 @@
...
+import glob
+import yaml
+from typing import List
+
+FOLDER_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+OUTPUT_PATH = os.path.join(os.path.dirname(__file__), "../README.md")
```
daabr: Better: use `pathlib`, and reuse the root directory.
Also "OUTPUT" is incorrect (we also read this file), and "FOLDER" is a little vague.

```suggestion
from pathlib import Path
from typing import List
import yaml


ROOT_PATH = pathlib.Path(__file__).parent
README_PATH = ROOT_PATH / "README.md"
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,137 @@
...
+                continue
+
+            new_lines.append(line)
+
+    with open(file_path, "w") as file:
+        file.writelines(new_lines)
+
+
+def main():
+    print("Starting README processing...")
```
daabr: Once you know this code works, I'd remove all these prints - they're not really useful in general: the user doesn't need to know that the program started, or finished, or exactly how many files were processed. The user only cares if there are errors/warnings.

scripts/create_project_table.py
```diff
@@ -0,0 +1,137 @@
...
+        print("No table header found.")
+        return
+
+    lines[header_index:header_index] = new_rows
+
+    with open(readme_file, "w") as file:
+        file.writelines(lines)
+
+
+def process_readme(file_path) -> None:
```
daabr: 1. "Process" is a little inaccurate, it implies some computation, "Reset" is more accurate
2. If you add a type hint (which is good), be consistent and add them everywhere (i.e. input parameters too)

scripts/create_project_table.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    Clean the README file by removing old table rows while keeping the header.
+
+    This prepares the README for a new table to be added.
+    """
```
daabr:
```suggestion
    """Reset the README.md file's table, in order to regenerate it."""
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,137 @@
...
+    new_lines = []
+    keep_table_header = False
+
+    with open(file_path, "r") as file:
```
daabr: When you use `pathlib` are suggested at the top, you can use https://docs.python.org/3/library/pathlib.html#reading-and-writing-files

Specifically, reading line-by-line is slow and consumes memory. It's quicker to read and write the entire file (https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text and https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text)

scripts/create_project_table.py
```diff
@@ -0,0 +1,137 @@
...
+        for line in file:
+            # Check if it's the table header or divider row
+            if line.strip().startswith("| Name") or line.strip().startswith("| :"):
```
daabr: Looking for the first and last lines is brittle. Much better: insert (invisible) comments into the README file, and use a multi-line regular expression to replace everything between them with an empty table.

daabr (commented):
First round of comments, let's continue offline.

Also, good function design - it's easy to write unit tests for them... So please add a unit tests file. :)

scripts/create_project_table.py
mario99logic: I don’t understand—where else should type hints be added?

scripts/create_project_table.py
mario99logic: I don’t understand—where else should type hints be added?

README.md
```diff
@@ -9,7 +9,7 @@ projects for:
 
 In addition, the [samples](./samples/) directory contains projects that
 demonstrate basic system features, integration APIs, and best practices.
-
+<!--start-table-->
```
daabr:
```suggestion

<!-- START-TABLE -->
```

README.md
```diff
@@ -32,7 +32,7 @@ demonstrate basic system features, integration APIs, and best practices.
 | 🐍 [Slack support](./slack_support/)                                             | Categorize Slack support requests using AI, and route them to the appropiate people      | Slack &harr; Gemini, Google Sheets          |
 | 🐍 [Task chain](./task_chain/)                                                   | Run a sequence of tasks with fault tolerance                                             | Slack                                       |
 | 🐍 [Webhook to Jira](./webhook_to_jira/)                                         | Create Jira issues based on HTTP GET/POST requests                                       | HTTP &rarr; Jira                            |
-
+<!--end-table-->
```
daabr:
```suggestion
<!-- END-TABLE -->

```

README.md
```diff
@@ -32,7 +32,7 @@ demonstrate basic system features, integration APIs, and best practices.
...
 > [!NOTE]
```
daabr: We don't need this note anymore.

scripts/create_project_table.py
```diff
@@ -0,0 +1,131 @@
+"""
+Process README files to extract metadata, generate a table of content, 
+and update a central README file with aggregated data.
+
+- Extract metadata from the top section of the README file for each project (e.g., between "---" delimiters).
+- Generate table rows summarizing the metadata
+- Insert the generated rows into the project table in the main README file
```
daabr: These steps are redundant - they basically repeat the same info as the summary with more detail, but that detail is not important at a high level.

scripts/create_project_table.py
```diff
@@ -0,0 +1,131 @@
...
+
+Classes:
```
daabr: No need to specify classes in the module docstring. In other words, keep just the summary as a single-line docstring.

scripts/create_project_table.py
```diff
@@ -0,0 +1,131 @@
...
+- ReadmeMetadataProcessor: Extracts and processes metadata from individual README files.
+"""
+
+import os
+import yaml
+from pathlib import Path
+from typing import List
```
daabr: No need to import `typing.List` - builtin `list` works too.

```suggestion
import os
from pathlib import Path
import yaml
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,131 @@
...
+
+    reset_readme(README_PATH)
+
+    new_rows = generate_readme_table(ROOT_PATH)
+
+    insert_rows_to_table(README_PATH, new_rows)
+
+
+if __name__ == "__main__":
+    main()
```
daabr:
```suggestion
if __name__ == "__main__":
    reset_readme(README_PATH)
    new_rows = generate_readme_table(ROOT_PATH)
    insert_rows_to_table(README_PATH, new_rows)
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,131 @@
...
+        # Check if it's the table header or divider row
+        if line.strip().startswith("| Name") or line.strip().startswith("| :"):
+            keep_table_header = True
+            new_lines.append(line)
+            continue
+
+        if keep_table_header and line.strip().startswith("|"):
+            continue
+
+        new_lines.append(line)
```
daabr: Much simpler than line-by-line: a multi-line regular expression.

In fact, I'm not sure this function is really needed. You can construct the table string in the final step and use a regular expression that replaces the entire table. 

```suggestion
    md = file_path.read_text(encoding="utf-8")
    headers = "-->\n| Name | Description | Integration |\n| :--- | :---------- | :---------- |\n<!--"
    md = re.sub("-->.+<!--", headers, md, flags=re.DOTALL)
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,131 @@
...
+import yaml
+from pathlib import Path
+from typing import List
+
+ROOT_PATH = Path(__file__).parent.parent
+README_PATH = ROOT_PATH / "README.md"
+
+
+class ReadmeMetadataProcessor:
+    def __init__(self, readme_file: Path):
```
daabr: Based on this `__init__` function, it looks like a class isn't really helping here a lot, there's no real state. You can just use module-scoped functions with parameters instead of class-scoped methods.

But feel free to wait a little with that, until we're done with the other comments.

scripts/create_project_table.py
```diff
@@ -0,0 +1,131 @@
...
+    rows = []
+    for readme_file in readme_files:
+        processor = ReadmeMetadataProcessor(readme_file)
+        row = processor.to_table_row()
+        if row:
+            rows.append(row)
+    return rows
+
+
+def insert_rows_to_table(readme_file: Path, new_rows: List[str]) -> None:
```
daabr:
```suggestion
def insert_rows_to_table(readme_file: Path, new_rows: list[str]) -> None:
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,131 @@
...
+
+        description = self.get_description()
+        integrations = self.get_integrations()
+        relative_path = os.path.relpath(os.path.dirname(self.readme_file), ROOT_PATH)
+        link = f"./{relative_path}/"
+
+        return f"| [{title}]({link}) | {description} | {integrations} |\n"
+
+
+def generate_readme_table(folder_path: Path) -> List[str]:
```
daabr:
```suggestion
def generate_readme_table(folder_path: Path) -> list[str]:
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,131 @@
...
+    """Generate a list of table rows from README metadata."""
+    readme_files = folder_path.rglob("README.md")
+
+    rows = []
+    for readme_file in readme_files:
```
daabr: No need for the helper var `readme_files` - the `rglob` call is short and clear on its own.

```suggestion
    rows = []
    for readme_file in folder_path.rglob("README.md"):
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,131 @@
...
+
+            metadata = yaml.safe_load("\n".join(metadata_lines))
+
+        return metadata
+
+    def get_title(self) -> str:
+        return self.metadata.get("title", None)
+
+    def get_description(self) -> str:
+        return self.metadata.get("description", "No description available.")
```
daabr: 1. Return `""` instead of `None` and `"No description available."
2. These function's aren't needed - `dict.get(key, "")` is a trivial call that doesn't need wrapping

scripts/create_project_table.py
```diff
@@ -0,0 +1,84 @@
+"""
+Process README files to extract metadata, generate a table of content, 
+and update a central README file with aggregated data.
+"""
```
daabr: Tiny cosmetic change. It's better to have a single line in the docstring, if possible.

```suggestion
"""Extract metadata from project README files to update table in repo's README file."""
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,84 @@
...
+        i for i, line in enumerate(lines) if line.strip() == "<!-- END-TABLE -->"
+    )
+
+    # Replace content between the markers
+    updated_lines = (
+        lines[: start_marker_index + 3]
+        + [line if line.endswith("\n") else f"{line}\n" for line in new_rows]
+        + lines[end_marker_index:]
+    )
+    readme_file.write_text("".join(updated_lines), encoding="utf-8")
```
daabr: There's no need for all this logic. It's simpler to just append the lines to `headers` before calling `md = re.sub()`:

```
for row in new_rows:
    table += row + "\n"

readme_file.write_text(table, encoding="utf-8")
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def to_table_row(readme_file: Path, metadata: dict) -> str:
+    """Convert metadata into a table row."""
+    title = metadata.get("title", "")
+    if not title:
+        return ""
+
+    description = metadata.get("description", "")
+    integrations = ", ".join(metadata.get("integrations", [])) or "None"
```
daabr: "None" is fine, but not really useful. Let's keep this value empty like the description if it's missing - it's easier to visually spot it visually that way.

scripts/create_project_table.py
```diff
@@ -0,0 +1,84 @@
...
+    relative_path = os.path.relpath(os.path.dirname(readme_file), ROOT_PATH)
+    link = f"./{relative_path}/"
+
+    return f"| [{title}]({link}) | {description} | {integrations} |\n"
```
daabr:
```suggestion
    path = os.path.relpath(os.path.dirname(readme_file), ROOT_PATH)

    return f"| [{title}](./{path}/) | {description} | {integrations} |\n"
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,84 @@
...
+                break
+            metadata_lines.append(line)
+
+        metadata = yaml.safe_load("\n".join(metadata_lines))
+
+    return metadata
+
+
+def to_table_row(readme_file: Path, metadata: dict) -> str:
+    """Convert metadata into a table row."""
```
daabr:
```suggestion
    """Convert metadata into a markdown table row."""
```

scripts/create_project_table.py
```diff
@@ -0,0 +1,84 @@
+"""
+Process README files to extract metadata, generate a table of content, 
+and update a central README file with aggregated data.
+"""
+
+import os
+from pathlib import Path
+import re
+import yaml
```
daabr: `yaml` is a non-builtin module. There's nothing bad with using it, but we don't really need it. See suggestion below.

scripts/create_project_table.py
```diff
@@ -0,0 +1,84 @@
...
+    if lines[0].strip() == "---":
+        metadata_lines = []
+        for line in lines[1:]:
+            if line.strip() == "---":
+                break
+            metadata_lines.append(line)
+
+        metadata = yaml.safe_load("\n".join(metadata_lines))
+
+    return metadata
```
daabr: Here's a simpler way to implement this without YAML.
Notice that I don't care if I add "fake" metadata fields from outside the header block, because `to_table_row` doesn't use them.

Also note that the suggestion below isn't complete: it doesn't handle lists (the integrations field) yet - but it's very close.

Also, unrelated to this PR, but a useful thing to learn: https://docs.python.org/3/library/dataclasses.html

```suggestion
    """Extract metadata from a project's README file."""
    field_pattern = r"^([a-z]+):\s+(.+)$"
    elements_pattern = r'"(.+?)"'
    f = readme_file.read_text(encoding="utf-8")
    metadata = {}

    for k, v in re.findall(field_pattern, f, re.MULTILINE):
        metadata[k] = v
        if k == "integrations":
            metadata[k] = ???

    return metadata
```

mario99logic:
where do I add the test file?, for now its in the same directory as the script file

scripts/tests/create_project_table_test.py
```diff
@@ -0,0 +1,139 @@
+"""
+Unit tests create_project_table.py.
+
+This suite tests functions for extracting metadata, generating markdown table rows, 
+processing multiple README files, and inserting rows into tables, using mocks to simulate file operations.
+"""
+
```
daabr: The documentation expectations for unit test modules are much more lax than regular code.
```suggestion
```

scripts/tests/create_project_table_test.py
```diff
@@ -0,0 +1,139 @@
...
+import unittest
+from unittest.mock import MagicMock, patch
+from pathlib import Path
+import sys
```
daabr: Maintain alphabetical order in imports

```suggestion
from pathlib import Path
import sys
import unittest
from unittest.mock import MagicMock, patch
```

scripts/tests/create_project_table_test.py
```diff
@@ -0,0 +1,139 @@
...
+from create_project_table import (
+    extract_metadata,
+    to_table_row,
+    generate_readme_table,
+    insert_rows_to_table,
+)
+
+
+class TestCreateProjectTable(unittest.TestCase):
+    def test_extract_metadata(self):
```
daabr:
```suggestion
class TestCreateProjectTable(unittest.TestCase):

    def test_extract_metadata(self):
```

scripts/tests/create_project_table_test.py
```diff
@@ -0,0 +1,139 @@
...
+            "integrations": ["slack", "gemini"],
+        }
+
+        expected_row = "| [Project One](./projects/project1/) | An example project for testing. | slack, gemini |\n"
+        row = to_table_row(mock_readme_file, metadata)
+        self.assertEqual(row, expected_row)
+
+    @patch("create_project_table.Path.rglob")
+    @patch("create_project_table.extract_metadata")
+    @patch("create_project_table.to_table_row")
```
daabr: Mocks are a powerful testing aid, but they're unnecessary and make testing harder and limited in this case. Let's discuss alternatives offline.

scripts/tests/create_project_table_test.py
```diff
@@ -0,0 +1,139 @@
...
+This suite tests functions for extracting metadata, generating markdown table rows, 
+processing multiple README files, and inserting rows into tables, using mocks to simulate file operations.
+"""
+
+import unittest
+from unittest.mock import MagicMock, patch
+from pathlib import Path
+import sys
+
+sys.path.append(str(Path(__file__).resolve().parent.parent))
```
daabr: This can violate test isolation. Tests should not affect the OS, and certainly not in a permanent way.

scripts/tests/create_project_table_test.py
```diff
@@ -0,0 +1,139 @@
...
+
+from create_project_table import (
```
daabr: - We do not import functions, we import modules (some exceptions apply, even in this PR, but not here)
- We do not group imports in a block, this is Python, not Go

daabr (commented):
Please move the file to the same directory as the script under test. No need for a separate `tests` subdirectory.

scripts/create_project_table_test.py
```diff
@@ -0,0 +1,124 @@
+from pathlib import Path
+import unittest
+from unittest.mock import MagicMock, patch
+import create_project_table
+
+
+class TestCreateProjectTable(unittest.TestCase):
+
```
daabr:
```suggestion
class TestCreateProjectTable(unittest.TestCase):
    """Unit tests of the "create_project_table" module."""

```

scripts/create_project_table_test.py
```diff
@@ -0,0 +1,124 @@
+from pathlib import Path
+import unittest
+from unittest.mock import MagicMock, patch
+import create_project_table
```
daabr:
```suggestion

import create_project_table
```

daabr (approved):

PR #126, opened 2024-12-07
docs: edit `categorize_emails` readme to favor cloud

pashafateev:


categorize_emails/README.md
```diff
@@ -9,64 +9,86 @@ categories: ["Office Automation"]
...
 ## How It Works
 
-- **Detect New Email**: The program monitors the Gmail inbox for new emails using the Gmail API.
-- **Categorize Email**: ChatGPT analyzes the email content and categorizes it into predefined categories.
-- **Send Slack Notification**: The program sends the categorized email content to the corresponding Slack channel using the Slack API.
-- **Label Email**: Adds a label to the processed email in Gmail for tracking.
+1. Fetches new emails from Gmail API
+2. Uses ChatGPT to analyze and categorize email content
+3. Posts categorized emails to relevant Slack channels
+4. Labels processed emails in Gmail
```
daabr: This is indeed more succinct and therefore better - good! My intuition is that a better tense for this list is to use "Fetch, Use, Post, Label" instead of "Fetches, Uses, Posts, Labels" - so it reads more like a recipe and less as an implicit reference to what "it" (the workflow) does. WDYT?

categorize_emails/README.md
```diff
@@ -9,64 +9,86 @@ categories: ["Office Automation"]
...
-### Configure integrations
+1. **Initialize Connections:** Initializing your connections through the UI.
+2. **Open Project Settings:** Navigate to the "TRIGGERS" tab in your project settings.
+3. **Edit Trigger:**
+   - Find the "receive_http_get" trigger under the "Actions" column.
+   - Click **Edit**.
+4. **Copy Webhook URL:** Copy the provided webhook URL.
+5. **Test with curl Command:** Test the webhook by sending a curl request with the webhook URL.
+      ```shell
+      curl -v "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}"
```
daabr: Let's standardize on using `-i` (include response headers in the output) instead of `-v` (verbose). "Verbose" is a little *too* verbose.

categorize_emails/README.md
```diff
@@ -9,64 +9,86 @@ categories: ["Office Automation"]
...
-- **Send Slack Notification**: The program sends the categorized email content to the corresponding Slack channel using the Slack API.
-- **Label Email**: Adds a label to the processed email in Gmail for tracking.
+1. Fetches new emails from Gmail API
+2. Uses ChatGPT to analyze and categorize email content
+3. Posts categorized emails to relevant Slack channels
+4. Labels processed emails in Gmail
 
 For more details, refer to [this blog post](https://autokitteh.com/technical-blog/from-inbox-to-slack-automating-email-categorization-and-notifications-with-ai/).
 
+## Cloud Usage (Recommended)
```
daabr, line 21: This is an important addition. I wonder (i.e. a preference, not a strict requirement) if you can make it more succinct like you did in "How it Works" above. For example:
- The highlighted summaries in steps 1, 3, 4, 5, 6 feel redundant
- The prefix in step 2 sounds like a separate step or simply not a highlighted summary, not a description of the end of the sentence

categorize_emails/README.md
```diff
@@ -9,64 +9,86 @@ categories: ["Office Automation"]
...
+> [!IMPORTANT]
+> Ensure all connections (Gmail, Slack, and ChatGPT) are properly initialized before the workflow starts running.
 
-1. Install and start a
-   [self-hosted AutoKitteh server](https://docs.autokitteh.com/get_started/quickstart),
-   or use AutoKitteh Cloud
+## Self-Hosted Deployment
 
-2. Run these commands to deploy this project's manifest file:
+#### Prerequisites
```
daabr: 1. Question: is it intentional to use here 4 "#" instead of 3? Do we display a table of contents in this page in which we want to hide this header?
2. Keep an empty line between headers and text above & below them.

categorize_emails/README.md
pashafateev: My concern with terms like “Fetch, Use, Post,” is that they might sound like instructions for the user, even though they are meant to describe what the workflow does. For example, if someone asks, “How does this workflow work?” you would respond with, “It fetches new emails, uses them for processing, and posts the results etc.”

That said, my primary concern is consistency throughout the documentation. Since I’m already reviewing all of it, this would be an ideal time to standardize.

categorize_emails/README.md
pashafateev: Done

categorize_emails/README.md
pashafateev: Yeah good call. Done

categorize_emails/README.md
pashafateev: 1. I thought visually it might look better, but I looked at it again and 3 makes more sense. Didn't understand about the table of contents and header. Could you elaborate?
2. Done

categorize_emails/README.md
daabr: I get your point, but I still prefer the base form over the third person form:

"Fetch" > "It fetches" ("It X..." sounds repetitive) > "Fetches".

categorize_emails/README.md
```diff
@@ -9,64 +9,88 @@ categories: ["Office Automation"]
...
 
-Run the following command, replacing {your-webhook-slug} with the webhook slug from the previous step:
+### Steps to Retrieve the Webhook URL
 
-```shell
-curl -v "http://localhost:9980/webhooks/{your-webhook-slug}"
-```
+- The webhook URL is provided in the output of the `ak deploy` command.
+- Run the following command to retrieve the URL:
+  ```bash
```
daabr: Shouldn't this be `shell`?

daabr (approved):
Approved with 2 tiny nits.

categorize_emails/README.md
pashafateev: Done

categorize_emails/README.md
pashafateev: Yes good catch. Done.

PR #134, opened 2024-12-19
fix(hubspot): skip users with errors

daabr:
Don't end the entire workflow if only a few users can't be added to Hubspot


pashafateev (approved):
do we want to keep this solution even after what we suspect is a python bug fix?

PR #136, opened 2024-12-19
fix: minor improvements

daabr:
1. Replace deprecated call
2. Reminder to replace `Exception` with a more specific error type

### DIFF
--- a/auth0_to_hubspot/program.py
+++ b/auth0_to_hubspot/program.py
@@ -38,10 +38,11 @@ def add_new_users(users):
         contact = _create_hubspot_contact(user)
         try:
             hubspot.crm.contacts.basic_api.create(contact)
-            print(f"Added {user['email']} to HubSpot")
+            print(f"Added to HubSpot: {user['email']}")
         except Exception as e:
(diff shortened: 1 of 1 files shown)


auth0_to_hubspot/program.py
```diff
@@ -39,8 +38,11 @@ def add_new_users(users):
         contact = _create_hubspot_contact(user)
         try:
             hubspot.crm.contacts.basic_api.create(contact)
-            print(f"Added {user['email']} to HubSpot")
+            print(f"Added to HubSpot: {user['email']}")
         except Exception as e:
+            # TODO: Replace "Exception" with a specific error for
+            # conflicts, i.e. don't ignore other types of errors.
+            # print(f"Contact already exists in HubSpot: {user['email']}")
```
pashafateev, line 45: what if it fails for a different reason other than "contact already exists"?

daabr: Right now, we still log but skip the error - after the future change, other failures (i.e. unexpected errors) will cause the workflow to fail, which is the general desired behavior

pashafateev (approved):

PR #137, opened 2024-12-23
feat: migrate Purrr to Python (part 1)

daabr:
This is a reasonable breakpoint in the migration from Starlark to Python: 1000 LoC is already a lot, and continuation is blocked by recent Python runtime issues.

Key things to review:
- Creating a new Slack channel when creating/reopening/undrafting a PR
- Slash commands: help, opt in/out
- "Infrastructure" in `slash_helper.py` and `users.py`
- A few new unit tests (might cover more functions in `slack_helper.py` with mocks)

FYI:
- New long-running workflow while for each active PR - to experiment with the viability of handling events and state there instead of asynchronously with Redis
- Preparing to migrate irreplaceable Redis functionalities with Google Sheets (not implemented yet, hopefully won't need to do a lot)

Refs: INT-65

### DIFF
--- a/purrr/github_pr.py
+++ b/purrr/github_pr.py
@@ -27,20 +27,20 @@ def _parse_github_pr_event(data) -> None:
     match data.action:
         # A new pull request was created.
         case "opened":
-            _on_pr_opened(data)
+            _on_pr_opened(data.action, data.pull_request)
         # A pull request was closed.
         case "closed":
             _on_pr_closed(data)
         # A previously closed pull request was reopened.
         case "reopened":
-            _on_pr_reopened(data)
+            _on_pr_reopened(data.action, data.pull_request)
 
         # A pull request was converted to a draft.
         case "converted_to_draft":
             _on_pr_converted_to_draft(data)
         # A draft pull request was marked as ready for review.
         case "ready_for_review":
-            _on_pr_ready_for_review(data)
+            _on_pr_ready_for_review(data.action, data.pull_request)
 
         # Review by a person or team was requested for a pull request.
         case "review_requested":
@@ -72,24 +72,24 @@ def _parse_github_pr_event(data) -> None:
         # - milestoned, demilestoned
 
 
-def _on_pr_opened(data) -> None:
+def _on_pr_opened(action: str, pr) -> None:
     """A new pull request was created (or reopened, or marked as ready for review).
 
     See also the functions "_on_pr_reopened" and "_on_pr_ready_for_review".
 
     Args:
-        data: GitHub event data.
+        action: GitHub PR event action.
+        pr: GitHub PR data.
     """
     # Ignore drafts until they're marked as ready for review.
-    if data.pull_request.draft:
+    if pr.draft:
         return
 
-    slack_channel.initialize_for_github_pr(data.action, data.pull_request)
+    slack_channel.initialize_for_github_pr(action, pr)
 
     # Keep an AutoKitteh session running as long as the PR is alive.
-    github_subs = autokitteh.subscribe(
-        "github_conn", filter="event_type == 'pull_request'"
-    )
+    filter = "event_type == 'pull_request' && data.number == " + pr.number
+    github_subs = autokitteh.subscribe("github_conn", filter=filter)
     while True:
         data = autokitteh.next_event([github_subs])
         print("Received GitHub PR event:", data.action)
@@ -116,7 +116,7 @@ def _on_pr_closed(data) -> None:
     pass  # TODO: Implement this function.
 
 
-def _on_pr_reopened(data) -> None:
+def _on_pr_reopened(action: str, pr) -> None:
     """A previously closed pull request (possibly a draft) was reopened.
 
     Slack bug alert from https://api.slack.com/methods/conversations.unarchive:
@@ -125,14 +125,15 @@ def _on_pr_reopened(data) -> None:
     rather than a bot token.
 
     Args:
-        data: GitHub event data.
+        action: GitHub PR event action.
+        pr: GitHub PR data.
     """
     # Ignore drafts - they don't have an active Slack channel anyway.
-    if data.pull_request.draft:
+    if pr.draft:
         return
 
     # Workaround for the Slack unarchive bug: treat this as a new PR.
-    _on_pr_opened(data)
+    _on_pr_opened(action, pr)
 
 
 def _on_pr_converted_to_draft(data) -> None:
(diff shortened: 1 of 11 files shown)


purrr/github_pr.py
```diff
@@ -0,0 +1,288 @@
...
+
+        # Ignored actions:
+        # - auto_merge_enabled, auto_merge_disabled
+        # - enqueued, dequeued
+        # - labeled, unlabeled
+        # - milestoned, demilestoned
+
+
+def _on_pr_opened(data) -> None:
+    """A new pull request was created (or reopened, or marked as ready for review).
```
pashafateev, line 76: is there a difference between "created" and "marked as ready for review" ?

purrr/slack_channel.py
```diff
@@ -0,0 +1,117 @@
...
+    error = "Failed to create Slack channel for " + pr.html_url
+    slack_helper.debug(error)
+    try:
+        slack.chat_postMessage(channel=user_id, text=error)
+    except SlackApiError as e:
+        error = f"Failed to notify <@{user_id}> (`{github_username}` in GitHub)"
+        error += " about failing to create Slack channel for "
+        error += f"{pr.html_url}: `{e.response["error"]}`"
+        slack_helper.debug(error)
+    return
```
pashafateev: The error handling here is impressively robust, with safeguards even within the handler itself. While that’s excellent for our internal use, we need to consider the user experience.

Purrrr stands out as the only project showcasing `ak`’s potential beyond basic workflows, which is crucial. However, my concern is that this PR has so many components it risks overwhelming users with its complexity.

For our workflows, we’ve generally adhered to a “less is more” principle. While Purrrr may be an exception, I’d still recommend simplifying it further if possible.

purrr/github_pr.py
```diff
@@ -0,0 +1,288 @@
...
+    github_subs = autokitteh.subscribe(
+        "github_conn", filter="event_type == 'pull_request'"
+    )
+    while True:
+        data = autokitteh.next_event([github_subs])
+        print("Received GitHub PR event:", data.action)
+        if data.action in ("closed", "converted_to_draft"):
+            _parse_github_pr_event(data)
+            autokitteh.unsubscribe(github_subs)
+            break
```
pashafateev, line 99: I could be mistaken, but it looks like any PR will unblock line 94. Is that the desired behavior?

If I open "PR 2" and then "PR 1" is closed and therefore unblocks, then when will we complete the cycle for "PR 2"?

purrr/slack_cmd.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Route further processing to the appropriate command handler.
+    if not args or "help" in args:
+        _help(data, args)
+        return
+
+    for cmd, func, _ in _COMMANDS:
+        if cmd == args[0]:
+            func(data, args)
+            return
```
pashafateev: nit: why not dictionary?

purrr/slack_cmd_test.py
```diff
@@ -0,0 +1,123 @@
+from datetime import datetime
```
pashafateev, line 1: nit : put test files in a separate test directory?

purrr/slack_cmd_test.py
```diff
@@ -0,0 +1,123 @@
...
+        event = autokitteh.AttrDict({"data": self.data | {"text": "help"}})
+        slack_cmd.on_slack_slash_command(event)
+
+        slack_cmd.slack.chat_postEphemeral.assert_called_once_with(
+            channel=event.data.channel_id,
+            user=event.data.user_id,
+            text=slack_cmd._help_text(event.data),
+        )
+
+    def test_on_slack_slash_command_with_noop_opt_in(self):
```
pashafateev, line 57: at this point is where the tests become a bit more complex and difficult for me to understand. Can we use this as an opportunity to teach users how to test their workflows?

i.e. provide some comments/docstrings

purrr/users.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Returns:
+        Slack user ID, or "" if not found.
+    """
+    # Optimization: if we already have it cached, no need to look it up.
+    slack_user_id = data_helper.cached_slack_user_id(github_username)
+    if slack_user_id:
+        if slack_user_id in ("bot", "not found"):
+            slack_user_id = ""
+        return slack_user_id
```
pashafateev:
```suggestion
    if slack_user_id and slack_user_id not in ("bot", "not found"):
        return slack_user_id
    return ""
```

purrr/github_pr.py
daabr: yes - "opened" is the event when you create a new PR or draft, "ready_for_review" is when you click "Ready for review" on an existing draft PR.

But in Purrr's case they're treated the same, because we ignore drafts.

purrr/slack_cmd_test.py
daabr: No, test files should be alongside the code that they're testing. Only test data file can/should be in their own subdirectory.

purrr/slack_cmd_test.py
daabr: AFAIK we don't really support unit testing in AK yet. You need a bit of trickery to run these tests, for now.

But I agree - we should bring this up, and treat unit tests as projects as a feature too - but not in this PR.

Filed INT-147 now.

purrr/github_pr.py
daabr: You're right! Fixed.

purrr/slack_cmd.py
daabr: Insertion order preservation has been an official feature in Python dictionaries only since https://docs.python.org/3.7/whatsnew/3.7.html, that's why I'm not used to it. But it's a good idea - done.

purrr/users.py
daabr: There's a bug in your suggestion: we never reach the code below it. But I simplified this block by removing the nesting.

purrr/slack_channel.py
daabr: Thanks, it's not **that** impressive... :)

Failures to handle opening a PR are unique compared to other events - they violate a user expectation that their PR is broadcasted, so they need to be notified reliably about it.

That said, I agree it's sufficient to let the workflow crash with an error (due to the Slack client's exception) if the notification itself fails - we can't work around that automatically, and workflow errors are a good signal and debugging tool for project/system issues, just like messages in the `purrr-debug` channel.

Also, after simplifying this code, I found a couple of small issues in it!

purrr/users.py
pashafateev: I think your change is easier to read, but I don't see the bug.

pashafateev (approved):

purrr/users.py
daabr: discussed offline

PR #139, opened 2024-12-29
feat(purrr): mention and impersonate GitHub users in Slack

daabr:
Also move the "debug" function from `slack_helper` to a new module to avoid circular imports.

Refs: INT-131

### DIFF
--- a/purrr/debug.py
+++ b/purrr/debug.py
@@ -6,6 +6,7 @@ Decoupled from the "slack_helper" module, to avoid circular imports.
 import os
 import re
 import traceback
+from typing import Iterator
 
 from autokitteh.slack import slack_client
 from slack_sdk.errors import SlackApiError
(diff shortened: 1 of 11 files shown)


purrr/debug.py
```diff
@@ -0,0 +1,50 @@
...
+    for file, line, func, text in traceback.extract_stack():
+        # Log only frame summaries relating to this project, up to this function.
+        if "/ak-user-" not in file:
+            continue
+
+        # Display shorter and cleaner paths.
+        file = re.sub(r"^.+/ak-user-.+?/", "", file)
+        msgs.append(f"```File: {file}, line {line}\nFunc: {func}\n{text}```")
+
+    return msgs[:-2]  # Skip the last 2 frames (i.e. this module).
```
pashafateev, line 50: these are for the time being? Didn't we discuss adding these changes to the runtime code?

purrr/markdown.py
pashafateev, line 1: I understand the need for this, but this is a perfect example of something that can be added as a library and then used across all workflows. Then these 2 files can be taken out of this directory.

purrr/users.py
```diff
@@ -26,6 +71,10 @@ def github_username_to_slack_user_id(github_username: str) -> str:
     Returns:
         Slack user ID, or "" if not found.
     """
+    # Don't even check GitHub teams, only individual users.
+    if "/" in github_username:
+        return ""
+
     # Optimization: if we already have it cached, no need to look it up.
     slack_user_id = data_helper.cached_slack_user_id(github_username)
```
pashafateev, line 161: This is not very likely, but say a user changes their username?
Do we just not support that?

purrr/users.py
```diff
@@ -67,30 +116,31 @@ def github_username_to_slack_user_id(github_username: str) -> str:
...
     # Optimization: cache unsuccessful results too (i.e. external users).
-    data_helper.debug(f"GitHub user {gh_user_link}: email & name not found in Slack")
+    debug.log(f"GitHub user {gh_user_link}: email & name not found in Slack")
     data_helper.cache_slack_user_id(github_username, "not found")
     return ""
 
 
-def _email_to_slack_user_id(email: str) -> str:
-    """Convert an email address to a Slack user ID.
+def resolve_github_user(github_user) -> str:
```
pashafateev: personally liked the old naming better. more clear as to what it's doing.

pashafateev (commented):
General question:
How do you decide what to include within a try/except block?

purrr/debug.py
daabr: No. `debug.log` may also be called without aborting the workflow, and this context would still be very useful.

purrr/markdown.py
daabr: agreed, i'll create a ticket to do this, maybe as a part of more generic "text utilities"?

purrr/users.py
daabr: I'm not sure a Slack user **can** change their username, only their profile details (e..g **display** name). Even if it is possible, I'm fine with the user having to opt-out again in this unusual situation, or we could create a "good first issue" to fix this.

purrr/users.py
daabr: This is just a GitHub mistake, `_email_to_slack_user_id` still exists - it just moved closer to the top. `resolve_github_user` is a new function that just happens to be in the same place.

daabr (commented):
> How do you decide what to include within a try/except block?

Usually it's a single operation. It's more than one if I'm ok with abandoning that sequence of steps half-way due to an error in any step.

purrr/markdown.py
pashafateev: yes I like that

purrr/users.py
pashafateev: I meant a github username. I don't think you can change a Slack username.

purrr/users.py
pashafateev:
```suggestion
def format_github_user_for_slack(github_user) -> str:
```

purrr/users.py
daabr: Yeah, you're right, but same answer: I'm fine with the user having to opt-out again, if they change their GitHub username. Not only is this rare, it's even less likely when the user is already a contributing member in a project/company.

purrr/markdown.py
daabr: https://linear.app/autokitteh/issue/INT-151/move-markdown-functions-from-purrr-to-aks-python-sdk

pashafateev (approved):

PR #140, opened 2025-01-01
feat(purrr): more basic functionalities

daabr:
- Intercept Slack events in PR workflow (experimental)
- Add Slack users to a channel when it's created
- Archive the Slack channel when the PR is closed or marked as a draft
- Add/remove reviewers

Refs: INT-152

### DIFF
--- a/purrr/github_pr.py
+++ b/purrr/github_pr.py
@@ -30,7 +30,7 @@ def _parse_github_pr_event(data) -> None:
             _on_pr_opened(data.action, data.pull_request, data.sender)
         # A pull request was closed.
         case "closed":
-            _on_pr_closed(data)
+            _on_pr_closed(data.action, data.pull_request, data.sender)
         # A previously closed pull request was reopened.
         case "reopened":
             _on_pr_reopened(data.action, data.pull_request, data.sender)
@@ -88,7 +88,7 @@ def _on_pr_opened(action: str, pr, sender) -> None:
 
     channel = slack_channel.initialize_for_github_pr(action, pr, sender)
 
-    # Keep an AutoKitteh session running as long as the PR is alive.
+    # Intercept relevant GitHub and Slack events.
     filter = f"event_type == 'pull_request' && data.number == {pr.number}"
     subs = [autokitteh.subscribe("github_conn", filter=filter)]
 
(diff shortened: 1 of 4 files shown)


purrr/github_pr.py
```diff
@@ -86,35 +89,53 @@ def _on_pr_opened(action: str, pr, sender) -> None:
...
+    # Intercept relevant GitHub and Slack events.
+    filter = f"event_type == 'pull_request' && data.number == {pr.number}"
+    subs = [autokitteh.subscribe("github_conn", filter=filter)]
+
+    filter = "(event_type == 'message' || event_type.startswith('member_'))"
+    filter += f" && data.event.channel == {channel}"
+    subs.append(autokitteh.subscribe("slack_conn", filter=filter))
+
+    filter = f"event_type == 'reaction_added' && data.event.item.channel == {channel}"
+    subs.append(autokitteh.subscribe("slack_conn", filter=filter))
```
pashafateev, line 103: help me understand the motivation here

purrr/github_pr.py
```diff
@@ -86,35 +89,53 @@ def _on_pr_opened(action: str, pr, sender) -> None:
...
+            print("Received GitHub PR event:", data.action)
+            if data.action in ("closed", "converted_to_draft"):
+                for sub in subs:
+                    autokitteh.unsubscribe(sub)
+                break
 
-def _on_pr_closed(data) -> None:
+        # Slack event.
+        else:
+            print("Received Slack event:", data.event.type)
```
pashafateev, line 119: you added subscription to slack events on line 100, but this is all you do with it. is that temporary?

purrr/github_pr.py
daabr: I want to try to handle some events(*) in this new long-running workflow instead of creating project-level triggers that start independent workflows. I'm still not sure if it'll work well, so it's experimental, and not actually doing anything for-realz yet.

`*` = Specifically syncing Slack activity to the PR in GitHub.

purrr/github_pr.py
daabr: See previous reply - it's experimental for now. I'll see if it can replace project triggers reliably as soon as ENG-1913 is fixed.

purrr/github_pr.py
pashafateev: My only comment then is if this stays, to document it in greater detail.

pashafateev (approved):

PR #144, opened 2025-01-04
refactor: copilot project

daabr:
Fixes in general:
- Use Copilot terminology from Copilot settings page: "cancel seat" instead of "un/subscribe" and "un/register" (except in the metadata description - where brevity is important)

Fixes in `README.md`:
- Removed ":" from a header
- Removed obsolete project var `GITHUB_ORG`
- Removed obsolete warning at the bottom, after https://github.com/autokitteh/docs.autokitteh.com/pull/143

Fixes in `autokitteh.yaml`:
- Removed obsolete project var `GITHUB_ORG`
- Renamed project vars and default Slack channel to be more self-describing
- `IDLE_USAGE_THRESHOLD`: changed debug values (10m) to stated default (72h)

Replaced `helpers.py` with `users.py` (adapted from new Purrr, like `helpers.py`)

Fixes in `users.py` compared to `helpers.py`:
- "Helpers" is too generic, always use specific names
- Never return `None` when a function declares that it returns only `str`
- When calling Slack functions, rely only in `try` & `except SlackApiError`, checking `resp.get("ok", False)` is never reached in case of an error (unlike our old Starlark wrapper)
- `_slack_users` function: Python rewrite doesn't need recursion (unlike our old Starlark wrapper)

Fixes in `msg.json`:
- UX principle: whenever possible, the "yes" option should be "first" (on the left), and the "no" option should be second (on the right)

Fixes in `seats.py`:
- Redundant constants and globals: if used only once, embedded them where they're used
- Use new project var names from the manifest (more descriptive)
- Lumped GitHub globals together
- `prune_idle_seats` function is unnecessary - embedded in `find_idle_seats` with a new input boolean
- `t` timestamp: renamed to be more self-describing (`now`), but more importantly eset for each seat instead of using a single initial value (the loop *should* be quick, but may not be)
- `status` line too line, split into 2 lines (readability is usually more important than efficiency)
- Reversed `is_idle` to `is_active`: results in flatter code at the bottom of the loop (`if is_active: continue` instead of `if is_idle: ...`)
- `seat_dict`: renamed, don't specify types in variable names
- `seat_dict`: changed type from `dict[str, Any]` to `dict[str, str]` - it's our object, no need to mimic GitHub's nested object structure
- Replaced "engage seat" (which doesn't mean a lot?) with "interact with user" in function name and comments
- Removed test-only comment and code
- Subscribe: use `event_type` instead of `data.type`
- Use more modern Python `match-case` instead of `if-elif-else`
- `report` function: handle gracefully empty `SLACK_LOG_CHANNEL` to enable making this project var optional

Note: we don't really need `autokitteh.start`, we could use the same trick as in the Slack sample project (embed important values e.g. GitHub/Slack user IDs in the interactive message and get them when the user clicks a button), but I guess there's educational value in showcasing `autokitteh.start`.

Fixes in `triggers.py`:
- `on_schedule` function: print each seat on its own instead of the whole list - more readable
- `_get_logins` function: renamed arg, don't specify types in variable names


github_copilot_seats/users.py
```diff
@@ -0,0 +1,69 @@
...
+    github_name = (user.name or "").lower()
+    if not github_name:
+        return ""
+
+    for user in _slack_users():
+        profile = user.get("profile", {})
+        real_name = profile.get("real_name", "").lower()
+        normalized_name = profile.get("real_name_normalized", "").lower()
+        if github_name in (real_name, normalized_name):
+            return user.id
```
daabr:
```suggestion
            return user.get("id", "")
```

github_copilot_seats/README.md
```diff
@@ -1,29 +1,33 @@
...
+# GitHub Copilot Seat Pruning
 
-This automation searches daily for all users in a GitHub organization that are actively using Copilot.
-If Copilot was not used in a preceding period, it automatically unregisters them, and then notifies them.
-Users can then optionally ask for their subscription to be reinstated.
+This automation enumerates once a day all the users in the GitHub organization
+that have access to [Copilot](https://github.com/features/copilot). If any of
+them haven't used it in a preceding period of time, it automatically marks
+their seat for cancellation in the next billing cycle, and notifies them in a
+Slack DM.
```
pashafateev:
```suggestion
This automation runs daily to list all the users in the GitHub organization who have access to [Copilot](https://github.com/features/copilot). If any of
them haven't used it in a preceding period of time, it marks
their seat for cancellation in the next billing cycle, and notifies them in a
Slack DM.
```

pashafateev: I think "enumerates" makes it unnecessarily difficult to understand. And I took out "automatically" because I think it's redundant.

github_copilot_seats/autokitteh.yaml
```diff
@@ -6,15 +6,13 @@ version: v1
...
-      name: LOGINS
+    - name: IDLE_HOURS_THRESHOLD
+      value: 72
+    - # Optional: manage GitHub Copilot subscriptions only for these users, separated by commas.
+      name: MANAGED_LOGINS
       value: ""
-    - name: LOG_CHANNEL
-      value: log
+    - name: SLACK_LOG_CHANNEL
+      value: copilot-log
```
pashafateev: `SLACK_LOG_CHANNEL` being pre-populated in my PR is an undesired remnant of testing . Are you sure we want to set this for the user?

github_copilot_seats/seats.py
```diff
@@ -1,117 +1,116 @@
...
-LOG_CHANNEL = os.getenv("LOG_CHANNEL")
+IDLE_HOURS_THRESHOLD = int(os.getenv("IDLE_HOURS_THRESHOLD", "72"))
+MANAGED_LOGINS = os.getenv("MANAGED_LOGINS")
 
-logins = LOGINS.split(",") if LOGINS else []
 github = github_client("github_conn")
-slack = slack_client("slack_conn")
-
-org = github.get_organization(ORG_NAME)
+org = github.get_organization(os.getenv("github_conn__target_name"))
```
pashafateev: My concern is that we’re prioritizing brevity at the expense of clarity. There’s no existing documentation for `github_conn__target_name`, and I only heard about it last week. Its name isn’t very intuitive. Can a user deduce that `github.get_organization` expects the organization name? Probably. Is breaking this process into two distinct steps technically redundant? Yes. However, making something that isn’t immediately clear more explicit is worth adding an extra line, in my opinion.

github_copilot_seats/seats.py
```diff
@@ -1,117 +1,116 @@
...
-        delta = t - seat.last_activity_at
-        is_idle = delta >= timedelta(minutes=IDLE_USAGE_THRESHOLD)
+        # Was the assigned seat being used recently?
+        now = datetime.now(UTC)
+        delta = now - seat.last_activity_at
+        is_active = delta < timedelta(hours=IDLE_HOURS_THRESHOLD)
 
-        comparison = ">=" if is_idle else "<"
-        status = f"{seat.assignee.login}: {t} - {seat.last_activity_at} = {delta} {comparison} {IDLE_USAGE_THRESHOLD} minutes"
+        comparison = "<" if is_active else ">="
```
pashafateev, line 43: Can you expand on your thought process behind switching from `is_idle` to `is_active`?

github_copilot_seats/seats.py
```diff
@@ -1,117 +1,116 @@
...
-def engage_seat(seat: dict[str, Any]) -> None:
-    """Engages a GitHub user assigned to a seat by identifying their corresponding Slack user and initiating a workflow.
+def prune_idle_seat(seat: dict[str, str]) -> None:
+    """Interacts via Slack with a GitHub user assigned to an idle Copilot seat.
 
-    Note:
-        This is designed to run as a child workflow using:
-        autokitteh.start(loc="seats.py:engage_seat", data={"seat": seat})
+    Note - this function runs as a child workflow, by calling:
+    autokitteh.start(loc="filename.py:function_name", data={...})
```
pashafateev: style preference but I think the other version is easier to read and visually looks cleaner.

github_copilot_seats/autokitteh.yaml
daabr: Now that it's optional, I see no problem in making it empty by default.

github_copilot_seats/seats.py
daabr: OK, brought back `GITHUB_ORG_NAME` const.

github_copilot_seats/seats.py
daabr: From the PR description:
> Reversed `is_idle` to `is_active`: results in flatter code at the bottom of the loop

```
if is_active:
  continue

block of code...
```

Vs.

```
if is_idle:
  block of code...
```

github_copilot_seats/seats.py
daabr: Done.

pashafateev (approved):

PR #146, opened 2025-01-05
ci: refresh settings and workflow

daabr:
- Enable unit testing (except in the Purrr project - those UT require AK context)
- Add various optional but important linters to Ruff
- Don't pin pytest and ruff versions, no point in updating it from time to time like in the AK repo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
-# Add "test" once we have unit tests
-all: deps format lint # test
+all: deps lint format test
 
 deps:
+	python -m pip install --upgrade pip
 	python -m pip install -r requirements.txt
 
 format: deps
@@ -11,6 +11,6 @@ lint: deps
 	ruff check --fix --output-format full .
 
 test: deps
-	pytest -v
+	pytest -v --ignore=purrr .
 
 .PHONY: all deps format lint test
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,22 @@
+[tool.ruff]
+src = ["./**"]
+target-version = "py312"
+
+[tool.ruff.lint]
+# https://docs.astral.sh/ruff/rules/
+extend-select = ["B", "C4", "D101", "D2", "D3", "DTZ", "E5", "I", "N", "PTH", "Q", "S", "UP", "W"]
+ignore = ["D213", "S113", "S311", "S314", "S4"]
+# TODO: BLE, D100, S113
+# MAYBE: D4, ANN, S314, FBT, CPY, TRY
+
+[tool.ruff.lint.isort]
+force-sort-within-sections = true
+lines-after-imports = 2
+order-by-type = false
+known-third-party = ["discord", "github"]
+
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
-# Add "test" once we have unit tests
-all: deps format lint # test
+all: deps lint format test
```
mario99logic, line 1: Is there a specific reason for changing the order between lint and format?

requirements.txt
```diff
@@ -1,2 +1,2 @@
-pytest ~= 8.2
-ruff ~= 0.5
+pytest
+ruff
```
mario99logic, line 2: Why did we remove the version constraint

daabr: As I wrote in the PR description: no point in updating them from time to time like in the AK repo, we're OK with (and in fact prefer) always using the latest release

Makefile
daabr: Yes. It turns out that import sorting happens in the lint step, we I'm allowing the formatter to react to that. Other than that the order doesn't really matter either way.

pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[tool.ruff]
+src = ["./**"]
+target-version = "py312"
+
+[tool.ruff.lint]
+# https://docs.astral.sh/ruff/rules/
+extend-select = ["B", "C4", "D101", "D2", "D3", "DTZ", "E5", "I", "N", "PTH", "Q", "S", "UP", "W"]
```
pashafateev, line 7: How would this work with the formatter?
> We recommend against using this rule alongside the [formatter](https://docs.astral.sh/ruff/formatter). The formatter enforces consistent indentation, making the rule redundant.

source https://docs.astral.sh/ruff/rules/tab-indentation/

my default formatter is Ruff for example

pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+ignore = ["D213", "S113", "S311", "S314", "S4"]
+# TODO: BLE, D100, S113
```
pashafateev, line 9:
```suggestion
# TODO: BLE, D100
```

Appears in both lines 8 and 9

pyproject.toml
pashafateev, line 1: my main concern is that we're adding too many rules at once

pyproject.toml
daabr: Yes, it means that S113 is currently disabled because it highlights existing issues in the code, but we want to fix them and prevent future issues.

pyproject.toml
daabr: I use Ruff too. All the relevant rules in that link (W191, D206, D300, Q000-003) are redundant but not problematic, so no need to disable.

pashafateev (approved):

pyproject.toml
daabr: discussed offline, specifically all the enabled linters are verified to work without issues that will disrupt our ongoing work.

PR #147, opened 2025-01-06
feat: Add update_projects_table check in CI 

mario99logic:
added two jobs to ci.yml:

- Run update_projects_table.py file
- Check if there are uncommitted changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Verify README.md is up to date
         run: |
           python update_projects_table.py
-         git diff --name-only --exit-code README.md
+          git diff --name-only --exit-code README.md
       # (Enable this step once we have unit tests)
       # - name: Test with pytest
       #   run: pytest -v
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,11 +27,10 @@ jobs:
         run: |
           ruff check --output-format github .
           ruff format --check .
-      - name: Run update_projects_table.py
-        run: python update_projects_table.py
-      - name: Verify no changes to README.md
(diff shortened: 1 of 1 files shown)


mario99logic:
Should I add a message to run update_projects_table manually if the check fails?

mario99logic:
Why do we require the user to run update_projects_table.py manually? Wouldn't it be better to run the script in CI and automatically commit the changes, or is it better for the user to run the script manually in case they want to make additional changes to the table?

.github/workflows/ci.yml
```diff
@@ -27,6 +27,11 @@ jobs:
         run: |
           ruff check --output-format github .
           ruff format --check .
+      - name: Run update_projects_table.py
+        run: python update_projects_table.py
+      - name: Verify no changes to README.md
+        run: |
```
daabr:
```suggestion
      - name: Verify README.md is up to date
        run: |
          python update_projects_table.py
```

.github/workflows/ci.yml
```diff
@@ -27,6 +27,11 @@ jobs:
...
+         git diff --name-only --exit-code README.md
```
daabr:
```suggestion
         git diff --name-only --exit-code README.md
```

daabr (approved):

.github/workflows/ci.yml
```diff
@@ -27,6 +27,10 @@ jobs:
         run: |
           ruff check --output-format github .
           ruff format --check .
+      - name: Verify README.md is up to date
+        run: |
+          python update_projects_table.py
+         git diff --name-only --exit-code README.md
```
daabr:
```suggestion
          git diff --name-only --exit-code README.md
```

PR #148, opened 2025-01-07
fix: typing strictness

daabr:
A few random fixes for type-related warnings from Pylance in VS Code, when type checking mode is "standard" instead of the default "off". Not specific to any project or task, and not a complete fix of the repo, just a few low-hanging fruits.

Most common: `str | None` --> `str` in places where it's clear that `None` is not expected.

```
os.getenv("var name") --> os.getenv("var name") or "default value"

dict.get("key") --> dict.get("key", "default value")
```

Also a few import tweaks to fix false-positive type warnings.

### DIFF
--- a/aws_health_to_slack/program.py
+++ b/aws_health_to_slack/program.py
@@ -8,7 +8,6 @@ https://aws.amazon.com/blogs/mt/tag/aws-health-api/
 from datetime import datetime, timedelta, UTC
 import json
 import os
-import re
 
 import autokitteh
 from autokitteh.aws import boto3_client
@@ -16,7 +15,7 @@ from autokitteh.google import google_id, google_sheets_client
 from autokitteh.slack import slack_client
 
 
-url = os.getenv("GOOGLE_SHEET_URL")
+URL = os.getenv("GOOGLE_SHEET_URL") or ""
 
 
 def on_schedule(_):
@@ -44,7 +43,7 @@ def on_schedule(_):
 def _read_google_sheet() -> dict[str, str]:
     """Read mapping of project tags to Slack channels from Google Sheet."""
     sheets = google_sheets_client("google_sheets_connection").spreadsheets().values()
-    rows = sheets.get(spreadsheetId=google_id(url), range="A:B").execute()
+    rows = sheets.get(spreadsheetId=google_id(URL), range="A:B").execute()
     return {row[0].strip(): row[1].strip() for row in rows.get("values", [])}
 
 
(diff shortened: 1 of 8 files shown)


aws_health_to_slack/program.py
```diff
@@ -59,7 +58,7 @@ def _aws_health_events() -> list[dict]:
     https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/health/client/describe_events_for_organization.html
     """
     try:
-        mins = int(re.match(r"(\d+)m", os.getenv("TRIGGER_INTERVAL")).group(1))
+        mins = int((os.getenv("TRIGGER_INTERVAL") or "1m")[:-1])
```
pashafateev, line 62:
```suggestion
        mins = int((os.getenv("TRIGGER_INTERVAL") or "1m")[:-1]) # Remove unit suffix and parse as int
```
Not attached to wording, but that wasn't obvious to me so I think a comment helps

data_pipeline/pipeline.py
```diff
@@ -65,13 +65,17 @@ def insert_records(db_dsn, records):
...
+            {
+                "track_id": track_id,
+                "n": i,
+                "lat": float(elem.get("lat", "0")),
+                "lng": float(elem.get("lon", "0")),
+                "height": float(elem.findtext(".//") or "0"),
+            }
+        )
+
+    return records
```
pashafateev, line 81: I don't understand either the original or modified pieces of code very well, but I guess that's not a comment on the change.

purrr/users.py
```diff
@@ -106,7 +106,11 @@ def format_slack_user_for_github(slack_user_id: str) -> str:
         debug.log(f"Slack user <@{slack_user_id}>: `real_name` not found in profile")
         return "Someone"
 
-    users = [user for user in _github_users() if user.name.lower() == slack_name]
+    users = []
+    for user in _github_users():
+        if user.name and user.name.lower() == slack_name:
+            users.append(user)
+
```
pashafateev, line 113: why the change? Is it so you can check `if user.name`?

pashafateev (approved):

data_pipeline/pipeline.py
daabr: The block in the middle is just a dictionary of parsed data - I changed it to ensure there are no exceptions if data is missing in the pre-parsed data: `get("key") -> get("key", "default")`.

Other than that, I reversed the top and bottom: so instead of a multi-line list comprehension (which is a readability anti-pattern) we now use a regular loop.

purrr/users.py
daabr: yes, and keeping that in a single-line comprehension makes the line too long, so I converted it to a regular loop.

PR #149, opened 2025-01-07
test(purrr): enable unit tests without AK

daabr:
Improve mocking, and expect dummy GitHub and Slack clients instead of `ConnectionInitError` when the AutoKitteh Python SDK is running within unit tests.

~This PR depends on https://github.com/autokitteh/autokitteh/pull/964 being merged, and a new AutoKitteh Python SDK version released in PyPI - which is why unit tests are currently failing.~

Also found and fixed a small bug in `slack_cmd.py`

Refs: INT-159

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
           ruff format --check .
 
       - name: Test with pytest
-        run: AUTOKITTEH_UNIT_TEST=1 pytest -v .
+        run: pytest -v .
 
       - name: Verify README.md is up to date
         run: |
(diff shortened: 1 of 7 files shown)


tebeka:
Since we're using pytest, I prefer to use function for tests and not `unittest.TestCase`
Also with mock, I prefer using [pytest.monkeypatch](https://docs.pytest.org/en/stable/how-to/monkeypatch.html)

daabr:
> Since we're using pytest, I prefer to use function for tests and not `unittest.TestCase` Also with mock, I prefer using [pytest.monkeypatch](https://docs.pytest.org/en/stable/how-to/monkeypatch.html)

Ack, I'll convert all UT in the repo to Pytest in a separate PR.

pashafateev (approved):

PR #150, opened 2025-01-08
feat: add hubspot sample

mario99logic:
Added a HubSpot sample with two basic functions, each triggered by a different webhook.

### DIFF
--- a/samples/hubspot/README.md
+++ b/samples/hubspot/README.md
@@ -22,55 +22,40 @@ API details:
 1. Create a new contact in HubSpot with predefined details.
 2. Retrieve and list all deals from HubSpot, displaying their IDs and names.
 
-## Deployment & Configuration
+## Cloud Usage (Recommended)
 
-### Cloud Usage
-
-- Initialize your connection with HubSpot through the UI
-
-### Self-Hosted Server
-
-#### Prerequisites
-
-- [Install AutoKitteh](https://docs.autokitteh.com/get_started/install)
-
-#### Installation Steps
-
-1. Clone the repository:
-   ```shell
-   git clone https://github.com/autokitteh/kittehub.git
-   cd kittehub/samples/hubspot
-   ```
-
-2. Start the AutoKitteh server:
-   ```shell
-   ak up --mode dev
-   ```
-
-3. Deploy the project:
-   ```shell
-   ak deploy --manifest autokitteh.yaml
-   ```
-
-   The output will show your connection IDs, which you'll need for the next step. Look for lines like:
-   ```shell
-   [exec] create_connection "hubspot/hubspot_conn": con_01jh2gx9jce6jr436fsz43g5zf created
-   ```
-   
-   In this example, `con_01jh2gx9jce6jr436fsz43g5zf` is the connection ID.
-
-4. Initialize your connections using the CLI:
-   ```shell
-   ak connection init <connection ID>
-   ```
+- Initialize your connection through the UI.
+2. Edit the trigger of the workflow you want to trigger, in the "TRIGGERS" tab, under the "Actions" column.
+3. Copy the provided webhook URL.
+4. Send the webhook with a curl GET or POST request, depending on the selected webhook.
+   - For create_contact_webhook:
+      ```shell
+         curl -X POST "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}?email=email&firstname=name&lastname=name"
+      ```
+   - For list_deals_webhook:
+      ```shell
+         curl -X GET "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}"
+      ```
 
 ## Trigger Workflow
 
-The workflow is triggered by sending an HTTP GET request.
+Trigger the workflow using the webhook URL. Refer to the [Cloud Usage](#cloud-usage-recommended) section for detailed steps.
+
+### Steps to Retrieve the Webhook URL
+
+- The webhook URL is provided in the output of the `ak deploy` command.
+- Run the following command to retrieve the URL:
+  ```shell
+  ak trigger get <trigger name or ID>
+  ```
 
 > [!TIP]
 > The workflow can also be triggered manually by clicking the "Run" button in the UI. Ensure the function you want to run is set as the entrypoint.
 
 > [!IMPORTANT]
 > Ensure that the connection with HubSpot is properly initialized before the workflow starts running.
+
+### Self-Hosted Server
+
+Follow these [detailed instructions](https://docs.autokitteh.com/get_started/deployment) to deploy the project on a self-hosted server.
 
--- a/samples/hubspot/README.md
+++ b/samples/hubspot/README.md
@@ -31,7 +31,10 @@ API details:
 
    - For create_contact_webhook:
       ```shell
-         curl -X POST "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}?email=email&firstname=name&lastname=name"
+      curl "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}" \
+           -d email=meow@autokitteh.com \
+           -d firstname=Kitty \
+           -d lastname=Meowington
       ```
    - For list_deals_webhook:
       ```shell
--- a/samples/hubspot/README.md
+++ b/samples/hubspot/README.md
@@ -19,8 +19,8 @@ API details:
 
 ## How It Works
 
-1. Create a new contact in HubSpot with predefined details.
-2. Retrieve and list all deals from HubSpot, displaying their IDs and names.
+1. Create a new contact in HubSpot with predefined details
+2. Retrieve and list all deals from HubSpot, displaying their IDs and names
 
 ## Cloud Usage (Recommended)
 
--- a/samples/hubspot/README.md
+++ b/samples/hubspot/README.md
@@ -54,10 +54,7 @@ Trigger the workflow using the webhook URL. Refer to the [Cloud Usage](#cloud-us
   ```
 
 > [!TIP]
-> The workflow can also be triggered manually by clicking the "Run" button in the UI. Ensure the function you want to run is set as the entrypoint.
-
-> [!IMPORTANT]
-> Ensure that the connection with HubSpot is properly initialized before the workflow starts running.
+> The list_deals workflow can also be triggered manually by clicking the "Run" button in the UI. Ensure list_deals is set as the entrypoint.
 
 ### Self-Hosted Server
 
--- a/samples/hubspot/README.md
+++ b/samples/hubspot/README.md
@@ -29,7 +29,7 @@ API details:
 3. Copy the provided webhook URL.
 4. Send GET or POST requests to the webhook of your choice:
 
-   - For create_contact_webhook:
+   - For `create_contact_webhook`:
       ```shell
       curl "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}" \
            -d email=meow@autokitteh.com \
--- a/samples/hubspot/README.md
+++ b/samples/hubspot/README.md
@@ -36,7 +36,7 @@ API details:
            -d firstname=Kitty \
            -d lastname=Meowington
       ```
-   - For list_deals_webhook:
+   - For `list_deals_webhook`:
       ```shell
          curl -X GET "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}"
       ```
--- a/samples/hubspot/README.md
+++ b/samples/hubspot/README.md
@@ -1,0 +1,76 @@
+---
+title: HubSpot
+description: Simple usage of the HubSpot API
+integrations: ["hubspot"]
+categories: ["Samples"]
+---
+
+# HubSpot Sample
+
+This [AutoKitteh](https://github.com/autokitteh/autokitteh) project
+demonstrates integration with [HubSpot](https://www.hubspot.com).
+
+It sends a couple of requests to the HubSpot API, adds a contact, retrieves deals, and prints the responses back to the user.
+
+API details:
+
+- [HubSpot API](https://pypi.org/project/hubspot-api-client/)
+- [Python client library](https://github.com/HubSpot/hubspot-api-python)
+
+## How It Works
+
+1. Create a new contact in HubSpot with predefined details.
+2. Retrieve and list all deals from HubSpot, displaying their IDs and names.
+
+## Deployment & Configuration
+
+### Cloud Usage
+
+- Initialize your connection with HubSpot through the UI
+
+### Self-Hosted Server
+
+#### Prerequisites
+
+- [Install AutoKitteh](https://docs.autokitteh.com/get_started/install)
+
+#### Installation Steps
+
+1. Clone the repository:
+   ```shell
+   git clone https://github.com/autokitteh/kittehub.git
+   cd kittehub/samples/hubspot
+   ```
+
+2. Start the AutoKitteh server:
+   ```shell
+   ak up --mode dev
+   ```
+
+3. Deploy the project:
+   ```shell
+   ak deploy --manifest autokitteh.yaml
+   ```
+
+   The output will show your connection IDs, which you'll need for the next step. Look for lines like:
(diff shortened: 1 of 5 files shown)


samples/hubspot/README.md
```diff
@@ -0,0 +1,76 @@
...
+   [exec] create_connection "hubspot/hubspot_conn": con_01jh2gx9jce6jr436fsz43g5zf created
+   ```
+   
+   In this example, `con_01jh2gx9jce6jr436fsz43g5zf` is the connection ID.
+
+4. Initialize your connections using the CLI:
+   ```shell
+   ak connection init <connection ID>
+   ```
+
```
pashafateev: Sorry, forgot to notify you but we're no longer keeping this section in the README's.
Take a look at [this sample](https://github.com/autokitteh/kittehub/blob/main/categorize_emails/README.md). We added a [generic page](https://docs.autokitteh.com/get_started/deployment) to our docs so we don't have to copy and paste so much text each time.

samples/hubspot/README.md
```diff
@@ -0,0 +1,76 @@
...
+## Trigger Workflow
+
+The workflow is triggered by sending an HTTP GET request.
+
+> [!TIP]
+> The workflow can also be triggered manually by clicking the "Run" button in the UI. Ensure the function you want to run is set as the entrypoint.
+
+> [!IMPORTANT]
+> Ensure that the connection with HubSpot is properly initialized before the workflow starts running.
+
```
pashafateev: if triggering the workflow is the same for both cloud and self-hosted, then put this section right after "#Cloud Usage"

samples/hubspot/program.py
```diff
@@ -0,0 +1,37 @@
...
+        contact_input = SimplePublicObjectInputForCreate(properties=contact_properties)
+
+        # Send a request to HubSpot to create the contact
+        response = hubspot.crm.contacts.basic_api.create(
+            simple_public_object_input_for_create=contact_input
+        )
+
+        print(f"Contact created with ID: {response.id}")
+    except ApiException as e:
+        print(e)
```
pashafateev: is the try except necessary here?

samples/hubspot/README.md
```diff
@@ -0,0 +1,76 @@
...
+1. Create a new contact in HubSpot with predefined details.
+2. Retrieve and list all deals from HubSpot, displaying their IDs and names.
+
+## Deployment & Configuration
+
+### Cloud Usage
+
+- Initialize your connection with HubSpot through the UI
+
```
pashafateev: What about deploy?

samples/hubspot/autokitteh.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - name: hubspot_conn
+    integration: hubspot
+  triggers:
+  - name: createWH
+    event_type: '*'
+    webhook: {}
+    call: program.py:create_contact
+  - name: listWH
+    event_type: '*'
+    webhook: {}
```
pashafateev: this isn't a comment on your code, this is a question for @daabr. I haven't seen this syntax before. Something new?

daabr: @pashafateev technically, I see it here: https://github.com/search?q=repo%3Aautokitteh%2Fautokitteh%20%22webhook%3A%20%7B%7D%22&type=code

But I don't know if it's current or even valid. Either way, we shouldn't use this in Kittehub.

@mario99logic please specify the standard trigger field `type: webhook` instead.

samples/hubspot/autokitteh.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+version: v1
+project:
+  name: hubspot
+  connections:
+  - name: hubspot_conn
+    integration: hubspot
+  triggers:
+  - name: createWH
+    event_type: '*'
```
daabr: For both `createWH` and `listWH`:
1. Style requirement: use snake case and no abbreviations, i.e. `create_contact_webhook` and `list_deals_webhook`
2. Don't accept all HTTP methods, be specific in the `event_type`. Creation of resources usually requires POST (even though you're not actually expecting any parameters, so GET is fine too), and retrieval usually requires GET

samples/hubspot/program.py
daabr: @mario99logic to expand on Pasha's question: if you're not actually handling the error, there's no point in catching it. In fact, it's better not to, so at least the user will see that the session ended with an error.

samples/hubspot/program.py
```diff
@@ -0,0 +1,38 @@
...
+from autokitteh.hubspot import hubspot_client
+from hubspot.crm.contacts import SimplePublicObjectInputForCreate
+from hubspot.crm.contacts.exceptions import ApiException
+
+
+hubspot = hubspot_client("hubspot_conn")
+
+
+def create_contact(event):
+    contact_properties = {
```
daabr, line 15: At least check if the event contains any of these parameters - if not, use these default values.

samples/hubspot/program.py
```diff
@@ -0,0 +1,38 @@
...
+        "email": "meow@autokitteh.com",
+        "firstname": "Kitty",
+        "lastname": "Meowington",
+    }
+    try:
+        contact_input = SimplePublicObjectInputForCreate(properties=contact_properties)
+
+        # Send a request to HubSpot to create the contact
```
daabr: 1. The code below this comment is trivial, so this comment is unnecessary, it doesn't add any information to the reader
2. If you would have kept this comment, I'd ask to add "." at the end

samples/hubspot/program.py
```diff
@@ -0,0 +1,38 @@
...
+            simple_public_object_input_for_create=contact_input
+        )
+
+        print(f"Contact created with ID: {response.id}")
+    except ApiException as e:
+        print(e)
+
+
+def list_deals(event):
+    # Retrieve all deals from HubSpot
```
daabr:
```suggestion
```

samples/hubspot/program.py
```diff
@@ -0,0 +1,38 @@
...
+    all_deals = hubspot.crm.deals.get_all()
+    for deal in all_deals:
+        print(f"Deal ID: {deal.id}, Deal Name: {deal.properties.get('dealname')}")
```
daabr:
```suggestion
    for deal in hubspot.crm.deals.get_all():
        print(f"Deal ID: {deal.id}, deal name: {deal.properties.get('dealname')}")
```

samples/hubspot/program.py
mario99logic: I added error handling based on the HubSpot API documentation to demonstrate that we can use the HubSpotException class in AK, it's not necessary. shall I remove it?

samples/hubspot/autokitteh.yaml
mario99logic: Its because we added the option to choose the event type of the webhook, I exported the manifest file from the UI after I made the sample and thats what it looks like.
I will use the old syntax

samples/hubspot/program.py
```diff
@@ -0,0 +1,35 @@
...
+def create_contact(event):
+    email = event.data.url.query.get("email", "meow@autokitteh.com")
+    firstname = event.data.url.query.get("firstname", "Kitty")
+    lastname = event.data.url.query.get("lastname", "Meowington")
+
+    contact_properties = {
+        "email": email,
+        "firstname": firstname,
+        "lastname": lastname,
+    }
```
daabr: Combine these 2 blocks together - there's no need for the helper vars.

Also note that POST requests usually carry data inside them, not in the URL. See my comments in the readme file.

```suggestion
    contact_properties = {
        "email": event.data.url.query.get("email", "meow@autokitteh.com"),
        "firstname": event.data.url.query.get("firstname", "Kitty"),
        "lastname": event.data.url.query.get("lastname", "Meowington"),
    }
```

samples/hubspot/program.py
```diff
@@ -0,0 +1,35 @@
+"""Demonstration of AutoKitteh's HubSpot integration.
+
+This script showcases two basic operations with the HubSpot API:
+creating a new contact and listing all deals.
+"""
+
+from autokitteh.hubspot import hubspot_client
+from hubspot.crm.contacts import SimplePublicObjectInputForCreate
```
daabr: It's preferable to import modules rather than specific classes/functions, unless there's a specific reason to do that.

So, here use:
```
from hubspot.crm import contacts
```
And inside `create_contact` use:
```
contacts.SimplePublicObjectInputForCreate
```

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+API details:
+
+- [HubSpot API](https://pypi.org/project/hubspot-api-client/)
+- [Python client library](https://github.com/HubSpot/hubspot-api-python)
+
+## How It Works
+
+1. Create a new contact in HubSpot with predefined details.
+2. Retrieve and list all deals from HubSpot, displaying their IDs and names.
```
daabr: List items are (usually) not considered full sentences
```suggestion
1. Create a new contact in HubSpot with predefined details
2. Retrieve and list all deals from HubSpot, displaying their IDs and names
```

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Cloud Usage (Recommended)
+
+- Initialize your connection through the UI.
```
daabr:
```suggestion
1. Initialize your connection through the UI
```

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+2. Edit the trigger of the workflow you want to trigger, in the "TRIGGERS" tab, under the "Actions" column.
+3. Copy the provided webhook URL.
+4. Send the webhook with a curl GET or POST request, depending on the selected webhook.
```
daabr: The webhook is the destination, you can't "send a webhook". Instead:
```suggestion
4. Send GET or POST requests to the webhook of your choice:

```

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+   - For create_contact_webhook:
+      ```shell
+         curl -X POST "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}?email=email&firstname=name&lastname=name"
```
daabr: 1. Align the content of the code block with the beginning 
of "```", not the end
2. Specify data as CLI flags, don't attach to the URL
3. Due to using `-d`, `POST` is already implied
```suggestion
      curl "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}" \
           -d email=meow@autokitteh.com \
           -d firstname=Kitty \
           -d lastname=Meowington
```

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+1. Create a new contact in HubSpot with predefined details.
+2. Retrieve and list all deals from HubSpot, displaying their IDs and names.
+
+## Cloud Usage (Recommended)
+
+- Initialize your connection through the UI.
+2. Edit the trigger of the workflow you want to trigger, in the "TRIGGERS" tab, under the "Actions" column.
+3. Copy the provided webhook URL.
+4. Send the webhook with a curl GET or POST request, depending on the selected webhook.
+   - For create_contact_webhook:
```
daabr:
```suggestion
   - For `create_contact_webhook`:
```

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+      ```shell
+         curl -X POST "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}?email=email&firstname=name&lastname=name"
+      ```
+   - For list_deals_webhook:
```
daabr:
```suggestion
   - For `list_deals_webhook`:
```

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+      ```shell
+         curl -X GET "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}"
```
daabr: `GET` is the default for simple requests
```suggestion
         curl "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}"
```

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+      ```
+
+## Trigger Workflow
+
+Trigger the workflow using the webhook URL. Refer to the [Cloud Usage](#cloud-usage-recommended) section for detailed steps.
```
daabr, line 46: Seems like it's cleaner and clearer to move step 4 to here.

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Steps to Retrieve the Webhook URL
```
daabr, line 48: "Steps to" is redundant text, it dilutes the point. I'm also not sure it belong **after** step 4, because you need to do it before step 4

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+- The webhook URL is provided in the output of the `ak deploy` command.
```
daabr, line 50: Actually, no, the CLI provides only the slug. Also, there's a UI way to get the URL - which does belong in the cloud usage section.

For a similar example, see the readme in #64 - but note that part of the instructions there because of the need to use a public host address in the URL, which is irrelevant in this project.

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+- Run the following command to retrieve the URL:
+  ```shell
+  ak trigger get <trigger name or ID>
+  ```
+
+> [!TIP]
```
daabr, line 56: Is this correct only for the list-deals webhook? In other words, even though the add-contact code has defaults in code, I'm not sure it can run with an empty event (like the one you get in a manual run)

samples/hubspot/README.md
```diff
@@ -0,0 +1,61 @@
...
+> The workflow can also be triggered manually by clicking the "Run" button in the UI. Ensure the function you want to run is set as the entrypoint.
+
+> [!IMPORTANT]
```
daabr: This note is correct, but personally I think it's not useful in a single-connection project which is focused on HubSpot anyway. It's kind of obvious.

samples/hubspot/README.md
```diff
@@ -0,0 +1,62 @@
...
+- The webhook URL is provided in the output of the `ak deploy` command.
+- Run the following command to retrieve the URL:
+  ```shell
+  ak trigger get <trigger name or ID>
+  ```
+
+> [!TIP]
+> The list_deals workflow can also be triggered manually by clicking the "Run" button in the UI. Ensure list_deals is set as the entrypoint.
+
+### Self-Hosted Server
```
daabr:
```suggestion
## Self-Hosted Server
```

samples/hubspot/program.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+hubspot = hubspot_client("hubspot_conn")
+
+
+def create_contact(event):
+    contact_properties = {
+        "email": event.data.url.query.get("email", "meow@autokitteh.com"),
+        "firstname": event.data.url.query.get("firstname", "Kitty"),
+        "lastname": event.data.url.query.get("lastname", "Meowington"),
```
daabr: You didn't test this after the readme update - these values are no longer in `event.data.url.query`, they are now in `event.data.body.form` - that's the whole point of using POST instead of GET: you pass data in the body of the request without limitations, instead of the URL.

daabr (approved):
Approved, but fix & test the comment in the Python code before merging.

samples/hubspot/program.py
mario99logic: I tested the setup, and it works when I run the following command:
curl -X POST "http://localhost:9980/webhooks/slug?email=newEmail@gmail.com"
I'm not sure what the issue is because when I tried using event.data.body.form, it didn't work. I’m confused about why the behavior is different.

mario99logic: I see what you mean now—when sending a POST request with data included in the body, it doesn’t seem to work as expected. 
because I did test it but I tested it with including the data in the URL

PR #151, opened 2025-01-09
test: convert all unit tests to Pytest

daabr:
Decided to use the `pytest-mock` plugin instead of the built-in `monkeypatch` which requires a lot more boilerplate in Purrr's tests.

Refs: INT-171

### DIFF
--- a/purrr/markdown_test.py
+++ b/purrr/markdown_test.py
@@ -1,276 +1,292 @@
 """Unit tests for the "markdown" module."""
 
 import collections
-import unittest
-from unittest import mock
 
 from autokitteh import github, slack
+import pytest
 
 
-# This is needed before calling "import markdown" to avoid "ConnectionInitError"
-# when initializing these clients in Purrr modules, due to the lack of AutoKitteh
-# environment variables. This is also the reason for the "noqa" comments below.
-github.github_client = mock.MagicMock()
-slack.slack_client = mock.MagicMock()
-
-import markdown  # noqa: E402
-import users  # noqa: E402
+@pytest.fixture(autouse=True)
+def setup_mock_github_and_slack_clients(mocker):
+    mocker.patch.object(github, "github_client", autospec=True)
+    mocker.patch.object(slack, "slack_client", autospec=True)
 
 
-class MarkdownGithubToSlackTest(unittest.TestCase):
+class TestGithubToSlack:
     """Unit tests for the "github_to_slack" function."""
 
     def test_trivial(self):
-        self.assertEqual(markdown.github_to_slack("", ""), "")
+        import markdown
 
-    def test_headers(self):
-        self.assertEqual(
-            markdown.github_to_slack("# Title 1\n\nFoo\n\n## Subtitle 2\nBar", ""),
-            "*Title 1*\n\nFoo\n\n*Subtitle 2*\nBar",
-        )
+        assert markdown.github_to_slack("", "") == ""
 
-    def test_text_style(self):
-        self.assertEqual(markdown.github_to_slack("_italic_", ""), "_italic_")
-        self.assertEqual(markdown.github_to_slack("*italic*", ""), "_italic_")
-        self.assertEqual(markdown.github_to_slack("__bold__", ""), "*bold*")
-        self.assertEqual(markdown.github_to_slack("**bold**", ""), "*bold*")
-        self.assertEqual(markdown.github_to_slack("***both***", ""), "_*both*_")
-        self.assertEqual(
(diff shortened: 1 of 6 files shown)


purrr/markdown_test.py
pashafateev, line 1: The main feedback here is similar to the comments on the original PR itself. `markdown_test.py` doesn’t seem specifically related to purrr. It feels more appropriate to place it in the `autokitteh` repository rather than `kittehub`.

daabr: Ack, in the separate PR. The unit tests will be the same either way.

purrr/slack_cmd_test.py
```diff
@@ -17,105 +20,120 @@
 }
 
 
-@mock.patch.object(slack_cmd, "slack", spec=WebClient)
-@mock.patch.object(slack_cmd, "data_helper", autospec=True)
-class SlackCmdTest(unittest.TestCase):
+# @mock.patch.object(slack_cmd, "data_helper", autospec=True)
+class TestSlackCmd:
```
tebeka: Why do we need this class? (vs top level test functions)

purrr/slack_cmd_test.py
```diff
@@ -17,105 +20,120 @@
 }
 
 
-@mock.patch.object(slack_cmd, "slack", spec=WebClient)
-@mock.patch.object(slack_cmd, "data_helper", autospec=True)
-class SlackCmdTest(unittest.TestCase):
+# @mock.patch.object(slack_cmd, "data_helper", autospec=True)
```
tebeka: Delete commented out code.

pyproject.toml
```diff
@@ -4,8 +4,8 @@ target-version = "py312"
 
 [tool.ruff.lint]
 # https://docs.astral.sh/ruff/rules/
-extend-select = ["B", "C4", "D101", "D2", "D3", "DTZ", "E5", "I", "N", "PTH", "Q", "S", "UP", "W"]
-ignore = ["D213", "S113", "S311", "S314", "S4"]
+extend-select = ["B", "C4", "D101", "D2", "D3", "DTZ", "E5", "I", "N", "PT", "PTH", "Q", "S", "UP", "W"]
+ignore = ["D213", "S101", "S113", "S311", "S314", "S4"]
 # TODO: BLE, D100, S113
```
tebeka: Is there an issue for these?

requirements.txt
```diff
@@ -1,4 +1,5 @@
 pytest
+pytest-mock
```
tebeka, line 2: Was aware of this one, cool

update_projects_table_test.py
```diff
@@ -1,119 +1,132 @@
...
-import tempfile
-import unittest
 
+import pytest
 import update_projects_table
 
 
-class TestCreateProjectTable(unittest.TestCase):
-    """Unit tests for the "update_projects_table" module."""
+class TestExtractMetadata:
```
tebeka, line 9: Why do we need the class? `self` is not used in any method.

update_projects_table_test.py
```diff
@@ -1,119 +1,132 @@
...
-    def generic_test_extract_metadata(self, input, expected):
-        actual = {}
-        with tempfile.NamedTemporaryFile(delete_on_close=False) as f:
-            if input:
-                f.write(input)
-            f.close()
-            actual = update_projects_table.extract_metadata(Path(f.name))
-        self.assertEqual(expected, actual)
+    @pytest.fixture
+    def tmp_file(self, tmp_path):
```
tebeka, line 13: I don't understand why do we need this fixture.

tebeka (approved):
Some comments, can commit.

purrr/slack_cmd_test.py
daabr: done

purrr/slack_cmd_test.py
daabr: you're right, I just modified the old `unittest`-based class. Done.

update_projects_table_test.py
daabr: clear visual separation between the tests of each function

update_projects_table_test.py
daabr: It's a Thin wrapper over `tmp_path` to create and delete a temporary file inside it. Not strictly required because `tmp_path` does its own cleanup, but I still think it's nicer.

pyproject.toml
daabr: Added INT-178, INT-179, INT-180.

PR #152, opened 2025-01-09
feat: new project to alert on missing events

daabr:
Tested manually with custom tweaks to use minute-based intervals and slack events instead of Jira.

Refs: INT-158

### DIFF
--- a/ops/alert_on_missing_events/README.md
+++ b/ops/alert_on_missing_events/README.md
@@ -1,0 +1,30 @@
+---
+title: Alert on missing events
+description: Send alerts when AutoKitteh doesn't receive certain events in time
+integrations: ["Slack"]
+categories: ["Ops"]
+---
+
+# Alert on Missing Events
+
+Send alerts when AutoKitteh doesn't receive certain events in time. This
+detects incidents due to missing updates from monitored services.
+
+Possible causes for example:
+
+- The monitored service is down / local network outage
+- The integration's callback URL has been modified
+- The connection's event watches have expired
+
+While an incident is ongoing, the workflow continues to wait for the desired
+events, and resends reminder alerts at a shorter interval.
+
+---
+
+You can add this project's configuration and code to existing projects, or run
+it in parallel to them. Either way, all matching triggers in AutoKitteh will
+receive each conformant event.
+
+You can also duplicate or extend this mechanism to handle multiple events and
+connections in a single project. AutoKitteh sessions are isolated from each
+other.
--- a/ops/alert_on_missing_events/README.md
+++ b/ops/alert_on_missing_events/README.md
@@ -1,6 +1,6 @@
 ---
-title: Alert on missing events
-description: Send alerts when AutoKitteh doesn't receive certain events in time
+title: Alert on missing Jira events
+description: Send Slack alerts when AutoKitteh doesn't receive certain Jira events in time
 integrations: ["Jira", "Slack"]
 categories: ["Ops"]
 ---
@@ -26,12 +26,11 @@ events, and resends reminder alerts at a shorter interval.
 1. Import/upload the project
 2. Initialize your connections
 
-> [!TIP] The default monitored service is Jira. If you want to monitor a
-> different service:
+> [!TIP] If you want to monitor a different service than Jira:
 >
 > 1. Delete the `monitored_service_conn` connection
-> 2. Recreate a new connection with the same name, and select the desired
->    integration
+> 2. Recreate a new connection with the same name,
+>    and select the desired integration
 > 3. Delete the `monitor_trigger` trigger
 > 4. Recreate a new trigger, select the new `monitored_service_conn`
 >    connection from step 2, and configure it
--- a/ops/alert_on_missing_events/README.md
+++ b/ops/alert_on_missing_events/README.md
@@ -7,10 +7,10 @@ categories: ["Ops"]
 
 # Alert on Missing Events
 
-Send alerts when AutoKitteh doesn't receive certain events in time. This is a
-detection tool for incidents due to missing updates from monitored services.
+Send alerts when AutoKitteh doesn't receive certain events in time.
 
-Possible causes for example:
+This is a detection tool for incidents due to missing updates from
+monitored services. Possible root-causes for example:
 
 - The monitored service is down / local network outage
 - The integration's callback URL has been modified
--- a/ops/alert_on_missing_events/README.md
+++ b/ops/alert_on_missing_events/README.md
@@ -51,7 +51,7 @@ events, and resends reminder alerts at a shorter interval.
      (default = `24` hours)
    - `PING_HOURS`: while an incident is ongoing, re-check at a shorter interval
      (default = `1` hour)
-   - `SLACK_CHANNEL`: send incident alerts on this Slack channel name/ID
+   - `SLACK_CHANNEL`: send incident alerts to this Slack channel name/ID
      (default = `autokitteh-alerts`)
 
 5. Deploy the project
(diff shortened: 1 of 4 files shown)


ops/alert_on_missing_events/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+> [!TIP] The default monitored service is Jira. If you want to monitor a
+> different service:
+>
+> 1. Delete the `monitored_service_conn` connection
+> 2. Recreate a new connection with the same name, and select the desired
+>    integration
+> 3. Delete the `monitor_trigger` trigger
+> 4. Recreate a new trigger, select the new `monitored_service_conn`
+>    connection from step 2, and configure it
```
pashafateev: This is not a comment on your instructions, which are fine, but rather a comment on the process itself. Are we satisfied with the level of effort required to make this change?

If the answer is yes—because changing existing connection details is not allowed—then I have a suggestion: Instead of assigning a default monitored service, could we leave the service blank by default? This would eliminate the need for an additional deletion step.

ops/alert_on_missing_events/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+3. Edit the trigger
+
+   - Specifically, select an `Event Type` and/or set the CEL expression in
+     the `Filter` field to match only the events you want to monitor
+
+4. Set/modify these project variables:
+
+   - `EVENT_FILTER`: must be identical/equivalent to the `Event Type` and/or
```
pashafateev, line 45: iiuc, `EVENT_FILTER` is required because of the `subscribe/next_event`. Is that critical for this demonstration? If it were event-driven, there wouldn’t be a need for a duplicate filter, which could introduce potential errors.

ops/alert_on_missing_events/README.md
```diff
@@ -0,0 +1,74 @@
...
+Follow [these detailed instructions](https://docs.autokitteh.com/get_started/deployment)
+to deploy the project on a self-hosted server.
+
+Also follow the instructions in the [Cloud Usage](#cloud-usage) section above.
+
+### Advanced Usage
+
+You can add this project's configuration and code to existing projects, or run
+it in parallel to them. Either way, all matching triggers in AutoKitteh will
+receive each conformant event.
```
pashafateev:
```suggestion
receive events that meet the specified criteria.
```
It's a loose suggestion. My issue is with the word "conformant". I had to look it up. Clarity > brevity here imo

ops/alert_on_missing_events/autokitteh.yaml
```diff
@@ -0,0 +1,47 @@
...
+    # While an incident is ongoing, re-check at a shorter interval.
+    - name: PING_HOURS
+      value: 1
+    # Send incident alerts to this Slack channel name/ID.
+    - name: SLACK_CHANNEL
+      value: autokitteh-alerts
+
+  connections:
+    # Set the "integration" field to the name of
+    # the integration that you want to monitor.
```
pashafateev: I understand the need for the other comments in this file, but I thought we were trying to comment less in the `.yaml`?
Are these important enough to repeat?

ops/alert_on_missing_events/program.py
```diff
@@ -0,0 +1,44 @@
...
+TIMEOUT_HOURS = int(os.getenv("TIMEOUT_HOURS", "24"))
+PING_HOURS = int(os.getenv("PING_HOURS", "1"))
+
+SLACK_CHANNEL = os.getenv("SLACK_CHANNEL", "")
+
+
+def on_monitor_trigger(event):
+    """Handle an incoming event from a monitored service."""
+    start_time = datetime.now(UTC)
+    slack = slack_client(CONN_NAME)
```
pashafateev, line 26: Why did you choose to initialize the client inside the function? I don’t think it’s a big deal, but I thought we were leaning toward initializing clients globally.

ops/alert_on_missing_events/program.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Wait for the next conformant event from the monitored service.
+    sub = autokitteh.subscribe(CONN_NAME, filter=EVENT_FILTER)
+    data = autokitteh.next_event(sub, timeout=timedelta(hours=TIMEOUT_HOURS))
+
+    # The monitored service hasn't sent us a conformant event for TIMEOUT_HOURS.
```
pashafateev, line 33: Same as above. I'm not sure what a conformant event is without looking it up

ops/alert_on_missing_events/program.py
```diff
@@ -0,0 +1,44 @@
...
+        description = EVENT_DESCRIPTION or f"`{EVENT_FILTER}` in `{CONN_NAME}`"
+        msg = f"Events not received since {start_time} (UTC): {description}"
+        slack.chat_postMessage(channel=SLACK_CHANNEL, text=msg)
+
+        data = autokitteh.next_event(sub, timeout=timedelta(hours=PING_HOURS))
+
+    # All clear, the monitored service is sending us conformant events still/again.
+    # Note that another "on_monitor_trigger" workflow is starting to run now, in a
+    # separate AutoKitteh session, waiting for the next event/incident.
+    autokitteh.unsubscribe(sub)
```
pashafateev, line 45: Question: why did you choose not to send any notification here that an event was received?

ops/alert_on_missing_events/README.md
daabr: 1. This tip is optional, so I'm fine with it being annoying
2. I'm not sure it's possible to define a connection without a type in the manifest? But I'll try - if that's possible, I also prefer that

ops/alert_on_missing_events/README.md
daabr: Yes, the point of this project is to wait for the **next** event, relying on a predefined trigger doesn't provide the desired functionality

ops/alert_on_missing_events/README.md
daabr: Agreed. Done with a tiny tweak

ops/alert_on_missing_events/program.py
daabr: Global client vars are useful when we use them multiple times and/or in multiple functions.

Both of these conditions don't happen here, so it's more wasteful than beneficial to make the client global.

Another note: global initializations happen unconditionally, and before the handler function is actually called. This is also wasteful here - in case we don't actually need to send an alert.

ops/alert_on_missing_events/program.py
daabr: Ack. I'm fine with a "technical" term in the code, as opposed to the readme.

ops/alert_on_missing_events/program.py
daabr: Yeah, I thought about it.

It'll require a small addition before the `while` loop (return if all is good before the loop, announce an "all-clear" only if we went through the loop and passed it.

WDYT? Would this workflow still be "short and sweet" with this change?

pashafateev: If receiving that message is important, I'd say it's worth it. I can't tell if it's important enough though.

ops/alert_on_missing_events/README.md
pashafateev: I would push back on the optional, unless this is going to be listed as "alert on missing jira events"

pashafateev (approved):

ops/alert_on_missing_events/autokitteh.yaml
daabr: done

ops/alert_on_missing_events/program.py
daabr: done

ops/alert_on_missing_events/README.md
daabr: done (added Jira, can't specify connection without type in manifest)

PR #157, opened 2025-01-11
refactor: self-describing project var

daabr:
Project vars: instead of `SLACK_CHANNEL`, use `SLACK_CHANNEL_NAME_OR_ID`. This will help cloud users, in the absence of documentation for project vars.

This is correct because all usages are in https://api.slack.com/methods/chat.postMessage, which accepts names too (which are easier to determine/set-up compared to IDs).

Detection: https://github.com/search?q=repo%3Aautokitteh%2Fkittehub%20SLACK_CHANNEL&type=code

### DIFF
--- a/hackernews/program.py
+++ b/hackernews/program.py
@@ -18,7 +18,7 @@ def on_slack_command(event):
     """Workflow's entry-point.
 
     Extracts a topic from a Slack command, monitors for new articles,
-    and posts updates to `SLACK_CHANNEL`.
+    and posts updates to the same Slack channel.
     """
     topic = event.data.text.split(" ", 1)[-1].strip()
     slack.chat_postMessage(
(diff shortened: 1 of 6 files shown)


hackernews/program.py
```diff
@@ -18,7 +18,7 @@ def on_slack_command(event):
     """Workflow's entry-point.
 
     Extracts a topic from a Slack command, monitors for new articles,
-    and posts updates to `SLACK_CHANNEL`.
+    and posts updates to the same Slack channel.
```
pashafateev, line 21: I can't tell what my stance on this change is, but just want to understand the reasoning

daabr: Simple: there is no `SLACK_CHANNEL` var in this specific project (in contrast to all the others), that's why I fixed it. I could write a much longer and more accurate description instead of "the same", but I opted for simplicity.

pashafateev: Oh lol. I just assumed there was – github didn't show me the whole file.

pashafateev (approved):

hackernews/program.py
mario99logic: I remember being confused when initializing this var first when I tried the slack sample, but I figured out that I needed to put the name of the channel. I think yes it's better to clarify this.

PR #162, opened 2025-01-15
feat(purrr): implement remaining PR events

daabr:
Un/assign reviewer, PR edited (title, description, base branch), PR synchronized (head branch).

Not tested manually yet (only unit testing) - this is still blocked by ENG-1913.

Refs: INT-152

### DIFF
--- a/purrr/github_pr.py
+++ b/purrr/github_pr.py
@@ -66,7 +66,7 @@ def _parse_github_pr_event(data) -> None:
             _on_pr_edited(data.action, data.pull_request, data.changes, data.sender)
         # A pull request's head branch was updated.
         case "synchronize":
-            _on_pr_synchronized(data)
+            _on_pr_synchronized(data.action, data.pull_request, data.sender)
 
         # TODO: locked, unlocked
 
@@ -370,6 +370,12 @@ def _on_pr_edited(action: str, pr, changes, sender) -> None:
     if not channel:
         return
 
+    # PR base branch was changed.
+    if "base" in changes:
+        msg = "{} changed the base branch from "
+        msg += "`{changes.base.ref}` to `{pr.base.ref}`"
+        slack_helper.mention_in_message(channel, sender, msg)
+
     # PR description was changed.
     if "body" in changes:
         if pr.body:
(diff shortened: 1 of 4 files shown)


purrr/github_pr.py
```diff
@@ -335,36 +348,79 @@ def _on_pr_unassigned(data) -> None:
...
+    if not channel:
         return
 
-    pass  # TODO: Implement this function.
+    assignee, sender = data.assignee, data.sender
+    _on_pr_review_request_removed_person(assignee, sender, channel, "assignee")
 
 
-def _on_pr_edited(data) -> None:
+def _on_pr_edited(action: str, pr, changes, sender) -> None:
```
mario99logic, line 359: Why not add type hints to the other parameters

purrr/github_pr.py
```diff
@@ -335,36 +348,79 @@ def _on_pr_unassigned(data) -> None:
...
+    channel = _lookup_channel(pr, action)
+    if not channel:
         return
 
-    pass  # TODO: Implement this function.
+    # PR base branch was changed.
+    if "base" in changes:
+        msg = "{} changed the base branch from "
+        msg += "`{changes.base.ref}` to `{pr.base.ref}`"
+        slack_helper.mention_in_message(channel, sender, msg)
```
mario99logic, line 377: Does base branch here mean the main branch and if so why are we checking if the main branch changed on a pr edited

mario99logic:
The tests effectively cover different scenarios for reviewers and assignees. Is there edge cases such a invalid user IDs?

mario99logic (approved):

purrr/github_pr_test.py
```diff
@@ -0,0 +1,192 @@
...
+def mock_slack_user_id(mocker):
+    import users
+
+    return mocker.patch.object(users, "github_username_to_slack_user_id", autospec=True)
+
+
+FakeGithubUser = collections.namedtuple("FakeGithubUser", ["login"])
+
+
+class TestOnPRReviewRequestedPerson:
```
mario99logic, line 25: The tests effectively cover different scenarios for reviewers and assignees. Is there edge cases such a invalid user IDs?

purrr/github_pr.py
daabr: There's no good definition of the structure of these objects, so no type hint is the most succinct way to show that. Another alternative is to specify an `Any` type, but that's redundant.

purrr/github_pr.py
daabr: `main` is a common base branch, but the PR author can use any branch they want, e.g. the head branch of another PR. A PR with a non-`main` base branch is called a "stacked PR" (i.e. it's "stacked" on top of another PR). You can read more about this here for example: https://axolo.co/blog/p/managing-stacked-pr

purrr/github_pr_test.py
daabr: Good question.
- The function under test (`_on_pr_review_requested_person`) receives its input from a GitHub event (that triggered the workflow)
- AK guarantees that this event is really from GitHub and not fake (we check signatures and discard fake payloads)
- So the `reviewer` and `sender` inputs are guaranteed to be valid GitHub user objects, specifically they contain a real `login` string field
- It **is** possible that a real GitHub user doesn't have a Slack account (e.g. external contributor, not an employee of the org)
- The function under test **does** handle this possibility, but it simply means returning without doing some things in Slack, so it's not an interesting test IMO

pashafateev (approved):
difficult to put this PR in perspective of the larger Purrr project, but nothing jumped out at me.

PR #164, opened 2025-01-16
feat(Auth0): Add auth0 sample

mario99logic:
Refs: INT-108


samples/auth0/README.md
```diff
@@ -0,0 +1,51 @@
...
+## How It Works
+
+1. Assign Roles to Users: Automatically assign roles to users based on their IDs.
+2. Fetch Weekly User Growth: Retrieve and display the number of new users created in the past week.
+
+## Cloud Usage 
+
+1. Initialize your connection with Auth0
+2. Copy the webhook URL from the "Triggers" tab (see the [instructions here](https://docs.autokitteh.com/get_started/deployment#webhook-urls))
+3. Update the `ROLE_ID` project variable with the `id` of the role you want to assign to the user
```
pashafateev: minimal change:
add an example of what a `ROLE_ID` looks like
```suggestion
3. Update the `ROLE_ID` project variable with the `id` of the role you want to assign to the user, e.g., `rol_fstvQ1eysK5VTkKt`
```

Should we also inform users that roles are created in their tenant dashboard, where they can find the role ID? This detail might not be obvious, especially to non-Auth0 users like myself. It also took me a while to find this information in the Auth0 documentation.

Thoughts @mario99logic @daabr ?

samples/auth0/README.md
```diff
@@ -0,0 +1,51 @@
...
+4. Deploy the project
+
+## Trigger Workflow
+
+`assign_role`:
+
+send an HTTP POST request to the the webhook URL from step 2 in the [Cloud Usage](#cloud-usage) section above:
```
pashafateev: nit:
```suggestion
Send an HTTP POST request to the the webhook URL from step 2 in the [Cloud Usage](#cloud-usage) section above:
```

samples/auth0/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+```shell
+curl -X POST "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}" \
+-d user_id=<user_id>
+```
+
+`weekly_user_growth`:
+
+workflow is automatically triggered every week.
```
pashafateev: nit:
```suggestion
Workflow is automatically triggered every week
```

samples/auth0/program.py
```diff
@@ -0,0 +1,30 @@
...
+
+def assign_role(event):
+    user = event.data.body.form.get("user_id")
+    auth0.roles.add_users(ROLE_ID, [user])
+    print(f"Assigned role {ROLE_ID} to user {user}")
+
+
+def weekly_user_growth(event):
+    """Fetch and display the number of users created in the past week."""
+    one_week_ago = (datetime.now(UTC) - timedelta(days=7)).isoformat()
```
pashafateev: Do we want to make this configurable? So make it a week by default, but make it easy to change with a constant?

samples/auth0/README.md
pashafateev: If we make the time configurable, then this would need to change. I would also add some information here about how to change the time for trigger.

samples/auth0/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Trigger Workflow
+
+`assign_role`:
+
+send an HTTP POST request to the the webhook URL from step 2 in the [Cloud Usage](#cloud-usage) section above:
+
+```shell
+curl -X POST "https://api.autokitteh.cloud/webhooks/{your-webhook-slug}" \
+-d user_id=<user_id>
```
daabr: 1. Add `-i` for more detailed output of `curl`
2. Don't use a specific hostname
3. After (2) `-d` can fit in the same line

```suggestion
curl -i -X POST "${WEBHOOK_URL}" -d user_id=${USER_ID}
```

samples/auth0/autokitteh.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - name: weekly
+    schedule: 0 0 * * 1
+    call: program.py:weekly_user_growth
+  - name: assign_role_webhook
+    type: webhook 
+    event_type: post
+    call: program.py:assign_role
+  vars:
+  - name: ROLE_ID
+    value: ""
```
daabr:
```suggestion
version: v1

project:
  name: auth0_sample

  vars:
    - name: ROLE_ID
      value: ""

  connections:
    - name: auth_conn
      integration: auth0

  triggers:
    - name: weekly
      schedule: 0 0 * * 1
      call: program.py:weekly_user_growth
    - name: assign_role_webhook
      type: webhook 
      event_type: post
      call: program.py:assign_role
```

samples/auth0/autokitteh.yaml
```diff
@@ -0,0 +1,20 @@
...
+# of an AutoKitteh project that demonstrates integration with Auth0.
+
+version: v1
+project:
+  name: auth0_sample
+  connections:
+  - name: auth_conn
+    integration: auth0
+  triggers:
+  - name: weekly
```
daabr: When using a cronspec, it's better to be clear about its meaning, i.e. name the trigger as "every_monday_at_midnight"

samples/auth0/program.py
```diff
@@ -0,0 +1,30 @@
...
+
+from datetime import datetime, timedelta, UTC
+import os
+
+from autokitteh.auth0 import auth0_client
+
+
+auth0 = auth0_client("auth_conn")
+
+ROLE_ID = os.getenv("ROLE_ID", "")
```
daabr: Reverse order - constants before variables.

samples/auth0/program.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def assign_role(event):
+    user = event.data.body.form.get("user_id")
```
daabr: Provide a default, even if it's `""`

samples/auth0/program.py
```diff
@@ -0,0 +1,30 @@
...
+
+from autokitteh.auth0 import auth0_client
+
+
+auth0 = auth0_client("auth_conn")
+
+ROLE_ID = os.getenv("ROLE_ID", "")
+
+
+def assign_role(event):
```
daabr, line 14: Document this function as a webhook-based workflow entry-point.

samples/auth0/program.py
```diff
@@ -0,0 +1,30 @@
...
+    user = event.data.body.form.get("user_id")
+    auth0.roles.add_users(ROLE_ID, [user])
+    print(f"Assigned role {ROLE_ID} to user {user}")
```
daabr:
```suggestion
    print(f"Assigned role {ROLE_ID!r} to user {user!r}")
```

samples/auth0/program.py
```diff
@@ -0,0 +1,30 @@
...
+
+def weekly_user_growth(event):
+    """Fetch and display the number of users created in the past week."""
+    one_week_ago = (datetime.now(UTC) - timedelta(days=7)).isoformat()
+    query = f"created_at:[{one_week_ago} TO *]"
+
+    response = auth0.users.list(q=query, search_engine="v3")
+    new_users = response.get("users", [])
+
+    print(f"New Users in the Past Week: {len(new_users)}")
```
daabr:
```suggestion
    print(f"New users in the past week: {len(new_users)}")
```

samples/auth0/program.py
```diff
@@ -0,0 +1,30 @@
...
+    for user in new_users:
+        print(f"- {user['email']} (Created At: {user['created_at']})")
```
daabr:
```suggestion
        print(f"- {user['email']} (created at: {user['created_at']})")
```

samples/auth0/program.py
```diff
@@ -0,0 +1,30 @@
...
+ROLE_ID = os.getenv("ROLE_ID", "")
+
+
+def assign_role(event):
+    user = event.data.body.form.get("user_id")
+    auth0.roles.add_users(ROLE_ID, [user])
+    print(f"Assigned role {ROLE_ID} to user {user}")
+
+
+def weekly_user_growth(event):
```
daabr: If you don't really need `event`, specify `_`
```suggestion
def weekly_user_growth(_):
```

samples/auth0/program.py
daabr: Expanding on this:
1. The trigger happens a few milliseconds after midnight, so `datetime.now(UTC)` is very close but not 100% accurate
2. Yes to @pashafateev's comment
3. As an example to solve this, see [this file](https://github.com/autokitteh/kittehub/pull/165/files#diff-563ff43f0458ef60d9a5a1bc30d5b209a0da5448a99221e3d1028d5e6a6c45ed) in PR 165 
4. See also the related changes in the README and YAML files - in your case you can use "d" (days) or "h" (hours) instead of "m" (minutes), I personally recommend 1h instead of 7d

samples/auth0/README.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Auth0
```
daabr, line 2:
```suggestion
title: Auth0 sample
```

mario99logic: But all of the titles in the  sample section is written without the word sample: asana, gemini, slack... why now use the word sample its inconsistent

daabr: See INT-183. But you're right, we can do that in a separate PR that fixes all projects in the samples directory.

samples/auth0/program.py
daabr: not fixed yet

samples/auth0/program.py
```diff
@@ -0,0 +1,34 @@
...
+    user = event.data.body.form.get("user_id", "")
+    auth0.roles.add_users(ROLE_ID, [user])
+    print(f"Assigned role {ROLE_ID!r} to user {user!r}")
+
+
+def weekly_user_growth(_):
+    """Fetch and display the number of users created in the past week."""
+    # Remove the unit suffix ("d") and parse as an integer.
+    interval_days = int((os.getenv("TIME_INTERVAL", "7d"))[:-1])
+    one_week_ago = (datetime.now(UTC) - timedelta(days=interval_days)).isoformat()
```
daabr: This is slightly inaccurate, see https://github.com/autokitteh/kittehub/blob/main/reliability/aws_health_monitor/program.py#L67-L69

samples/auth0/program.py
```diff
@@ -0,0 +1,34 @@
...
+
+    query = f"created_at:[{one_week_ago} TO *]"
```
daabr: This is slightly inaccurate, use `TO {now.isoformat()}]"` (where seconds and microseconds are reset, to ignore users who were added right now)

mario99logic: I rounded to the minute `end_time = datetime.now(UTC).replace(second=0, microsecond=0)` like in this example https://github.com/autokitteh/kittehub/blob/main/reliability/aws_health_monitor/program.py#L67-L69

mario99logic:
added     start_time_iso = start_time.isoformat().replace("+00:00", "Z")
    end_time_iso = end_time.isoformat().replace("+00:00", "Z") to match the format for the auth0 function

samples/auth0/program.py
```diff
@@ -0,0 +1,38 @@
...
+    """Entry-point function for a webhook-based workflow assigning a role to a user."""
+    user = event.data.body.form.get("user_id", "")
+    auth0.roles.add_users(ROLE_ID, [user])
+    print(f"Assigned role {ROLE_ID!r} to user {user!r}")
+
+
+def weekly_user_growth(_):
+    """Fetch and display the number of users created in the past week."""
+    # Remove the unit suffix ("d") and parse as an integer.
+    interval_days = int((os.getenv("TIME_INTERVAL", "7d"))[:-1])
```
pashafateev, line 24: What's the motivation behind using "7d"? Is it to make the unit of time clear? If so, couldn't we use a comment in the manifest or rename the variable to something like `DAYS_INTERVAL`?

mario99logic: its just a default value if, Daniel said to always add a default value, I can initialize `TIME_INTERVAL` to be 7d instead of putting a default but I think it doesnt really matter in my opinion

samples/auth0/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+```shell
+curl -i -X POST "${WEBHOOK_URL}" -d user_id=${USER_ID}
+```
+
+`weekly_user_growth`:
+
+Workflow is automatically triggered every week
+
+[!TIP]
```
daabr:
```suggestion
> [!TIP]
```

samples/auth0/program.py
daabr: I think @pashafateev's question was about the `"d"` and not about the existence of the default value.

The answer is that this is adapted from other projects, e.g. https://github.com/autokitteh/kittehub/blob/70ca702667a544a2ff4919f601673ba9f187754e/reliability/session_errors_monitor/autokitteh.yaml#L28

The point of the unit there is that the entire value (`"1m"`) is identical to the value in a schedule trigger (`"@every 1m"`).

In contrast, this project uses a cronspec string (`0 0 * * 1`).

@mario99logic maybe change the schedule in the manifest to be `"@every 168h"`, and also use "168h" here? This would solve Pasha's question, and also make the schedule's definition easier to understand.

Why not "7d"? According to Temporal's documentation, `@every X` parses `X` using https://pkg.go.dev/time#ParseDuration, which accepts `h` as the largest unit.

But this is just a suggestions, because "168h" is not as clear as "7d". I'm fine with this either way.

daabr (approved):

PR #166, opened 2025-01-17
refactor: GitHub actions project

daabr:
1. Move the "GitHub Action" project to the "devops" directory, and rename to "GitHub Workflows"
2. Refactor the project to be more user friendly (README + cleaner code)
3. Deleted the GitHub sample project - it's minimalistic, confusing and clunky, and obsoleted by this and other projects in Kittehub already

Tested manually in a local server with the internal tests repo, and the 2 dispatch workflows there.

Refs: INT-182


samples/github/README.md
pashafateev, line 1: I'm fine with deleting this particular github sample, but I still think that we should have a simple example for each integration. One option would be to delete the `workflow.py` file. Or to add a different sample.

pyproject.toml
pashafateev, line 1: not familiar with the purpose of this file in the kittehub repo. something new?

samples/github/README.md
daabr: Discussed offline. Keeping a README file with links to relevant projects elsewhere in the repo. For now, without metadata, so it won't be picked up by the web UI, because the DevOps category is discoverable on its own like the Samples category.

pyproject.toml
daabr: This contains the configuration for Ruff auto-formatting - specifically the change here means we don't have to force Ruff to know that `"import github"` means PyGithub (a third-party) instead of our first-party GitHub sample project, because there is no more first-party GitHub sample project.

daabr: anyway, I had to add it back

devops/github_workflows/program.py
```diff
@@ -0,0 +1,167 @@
...
+        completed_workflows.add(path)
+
+    _dispatch_workflow(C)
+
+
+def long_sequence(_) -> None:
+    """Long sequence demo (A --> B --> C --> A --> B --> C).
+
+    This is useful even within a single repository, because GitHub's own
+    workflow orchestration is limited to chaining no more than 4 workflows:
```
pashafateev:
```suggestion
    This is useful even within a single repository because GitHub workflows cannot chain more than 4 workflows using `workflow_run`:
```

pashafateev (approved):

PR #171, opened 2025-01-20
script for ad-hoc stats about projects

daabr:


list_projects_by.py
```diff
@@ -0,0 +1,28 @@
...
+import collections
+import sys
+
+from update_projects_table import extract_metadata
+from update_projects_table import ROOT_PATH
+
+
+if __name__ == "__main__":
+    if len(sys.argv) != 2 or sys.argv[1] not in ("category", "integration"):
+        print(f"Usage: {sys.argv[0]} <category|integration>")
```
mario99logic: I noticed an issue with the labels. The code doesn’t print anything when using "category" or "integration" because the metadata keys are actually "categories" and "integrations". When I updated the code to use "categories", it worked as expected. Also, in the "cat cafe" template, the correct keys are "integrations" and "categories".

list_projects_by.py
```diff
@@ -0,0 +1,28 @@
...
+        sys.exit(1)
+
+    label = sys.argv[1]
+    data = collections.defaultdict(list)
+    for readme_file in sorted(ROOT_PATH.rglob("README.md")):
+        metadata = extract_metadata(readme_file)
+        if metadata:
+            metadata["path"] = readme_file.parent.relative_to(ROOT_PATH)
```
mario99logic: It's a bit unclear to add a new key like path directly within this script. It might be cleaner to handle this in a function.

update_projects_table.py
```diff
@@ -56,5 +47,9 @@ def insert_rows_to_table(readme_file: Path, new_rows: list[str]) -> None:
 
 
 if __name__ == "__main__":
-    new_rows = generate_readme_table(ROOT_PATH)
-    insert_rows_to_table(ROOT_PATH / "README.md", new_rows)
```
mario99logic, line 60: Why not keep generate_readme_table as a standalone function dedicated solely to generating the README content?

mario99logic (commented):
Nice additions overall, Just a couple of notes.

list_projects_by.py
pashafateev: I agree. It's hard to understand

update_projects_table.py
```diff
@@ -7,24 +7,15 @@
...
-    return [row for row in rows if row]  # Remove empty rows.
-
-
 def extract_metadata(readme_file: Path) -> dict:
     """Extract metadata from a project's README file."""
     field_pattern = r"^([a-z]+):\s+(.+)"  # "key: value"
     f = readme_file.read_text(encoding="utf-8")
     metadata = {}
 
+    list_values = ("categories", "integrations")  # Others are just strings.
```
pashafateev: > \# Others are just strings.

I don't understand this comment.

list_projects_by.py
daabr: oops, you're right, fixed

list_projects_by.py
daabr: I think the hard part was `if metadata`. Actually, everything works correctly without it, so I removed it.

If I were to split this into a separate function it would look like this:

```py
def extract_metadata_with_path(readme_file: Path) -> dict:
    """Thin wrapper that adds the project's relative path to its metadata."""
    metadata = extract_metadata(readme_file)
    metadata["path"] = readme_file.parent.relative_to(ROOT_PATH)
    return metadata
```

As you can see, there's really no value in having this function just for the sake of calling `metadata["path"] = ...`.

update_projects_table.py
daabr: Notice that we didn't add unit tests for it. The lack of UT isn't the reason, but it's a sign that it's just a simple combination of other functions, it doesn't have interesting logic that's worth testing.

That and the fact that it's short and trivial make it natural to just put it inside `__main__`, which is supposed to be trivial and isn't testable by definition.

update_projects_table.py
daabr: improved

pashafateev (approved):

PR #181, opened 2025-01-23
fix: jira sample Readme 

mario99logic:


samples/atlassian/jira/README.md
```diff
@@ -25,37 +25,23 @@ Python code samples:
...
-   Please run this to complete:
+> [!NOTE]
+> For more details on connection options and setup, visit [this guide](https://docs.autokitteh.com/integrations/atlassian/connection).
 
-   ak connection init <connection ID>
-   ```
+## Trigger Workflow
 
-3. Create a new issue in Jira, and check its comments
+- Triggered by creating a new issue.
```
daabr: 1. This section is trivial and obvious, it's just a repeat of step 1 in "How It Works" - so I think it's better to just get rid of it in this project
2. If you want to keep it anyway:
   1. it's not an absolute or formulaic rule, but in general it's better to write in an active voice ("do X") instead of passive ("X is done"), e.g. "Create a new issue in Jira to trigger this project."
   2. If there's just a single step, there's no need to point in using a list

daabr (approved):
Approved with 1 comment - choose which option you prefer

samples/atlassian/jira/README.md
pashafateev: @daabr I would argue that the "How It Works" section is irrelevant. This sample has two different entry points and while not terribly complex, I think it could be beneficial to describe that.

So in the very least I think this is missing another way to trigger the workflow, by commenting on a Jira issue, but I would take out the "How It Works" section as well.

PR #186, opened 2025-01-26
feat: Add convention tests for metadata

mario99logic:
Added convention tests for readme metadata, for any If you have more ideas for conventions to enforce, discuss with me.

refs: INT-183 (title of the branch is 182, it should be 183)


mario99logic:
@daabr I have a question you said projects 1-3 covers all projects under the root directory. there are readmes that do not belong to any project how do I avoid them in the first test

tests/metadata_definitions.py
```diff
@@ -0,0 +1,42 @@
+"""Defines metadata schema, allowed integrations, and categories.
+
+- METADATA: Required metadata fields in README files.
+- INTEGRATIONS: List of allowed integration values.
+- CATEGORIES: List of allowed category values.
+"""
```
daabr: This docstring isn't bad, but there's no non-obvious information here, so you might as well skip it

tests/metadata_definitions.py
```diff
@@ -0,0 +1,42 @@
...
+
+METADATA = ["title", "description", "integrations", "categories"]
+
+INTEGRATIONS = [
```
daabr: 1. `INTEGRATIONS` is fine, but `ALLOWED_INTEGRATIONS` is even better
2. Same in `CATEGORIES`
3. Sort all the values in `INTEGRATIONS` and `CATEGORIES` alphabetically, for easy maintenance in the future
4. List are mutable, tuples are not; constants should be immutable for safety - so change `[]` to `()`

update_projects_table.py
```diff
@@ -23,8 +23,11 @@ def extract_metadata(readme_file: Path) -> dict:
     metadata = {}
 
     for k, v in re.findall(field_pattern, f, re.MULTILINE):
-        # Integrations value is a list of strings, others are just strings.
-        metadata[k] = re.findall(r'"(.+?)"', v) if k == "integrations" else v
+        # Integrations and categories are a list of strings, others are just strings.
+        if k in {"integrations", "categories"}:
```
daabr: It's an interesting decision to use a set here:

- Good: O(1) search time, compared to a tuple's O(N) - but there are only 2 elements, so performance doesn't matter
- Bad: mutable type, compared to a tuple - but it's anonymous so no one can change it anyway
- Bad: sets take more memory than a tuple, comparable to a dictionary

Overall, it's a good intuition, but I slightly prefer a tuple in this specific case

tests/convention_test.py
```diff
@@ -0,0 +1,74 @@
...
+    (Path(ROOT_PATH / "samples"), "Samples"),
+]
+
+METADATA_VALID = [
+    (metadata_definitions.CATEGORIES, "categories"),
+    (metadata_definitions.INTEGRATIONS, "integrations"),
+]
+
+
+def test_metadata_block_exists():
```
daabr: As written in INT-186: "[run these tests] if a metadata block exists in a readme file". In other words, don't assume all readme files **should** have a metadata block, test only the ones that have it.

We want to allow projects to to be excluded from the repo README and the web UI's import options.

tests/convention_test.py
```diff
@@ -0,0 +1,74 @@
...
+    """Checks for required metadata in README files."""
+    for readme_file in sorted(Path(ROOT_PATH).rglob("README.md")):
+        metadata = update_projects_table.extract_metadata(readme_file)
+
+        assert metadata, f"Metadata block is missing in the README file: {readme_file}"
+        assert all(key in metadata for key in metadata_definitions.METADATA), (
```
daabr: The first assert is wrong, as noted above.

As for asserts 2-4:
1. Should be in separate test functions because they're not related to each other; we don't have an error in assert 2 to hide additional errors in asserts 3 and 4
2. The relative path of a project is more useful than its title, to know where the fix should be, also it's easier for the eye if there's a visual separation, like:
   ```py
   error = "blah blah blah: /dir/subdir/README.md"
   ```

tests/convention_test.py
```diff
@@ -0,0 +1,74 @@
...
+        )
+
+
+@pytest.mark.parametrize(("allow_list", "metadata_type"), METADATA_VALID)
+def test_metadata_valid(allow_list, metadata_type):
+    """Checks that metadata fields contain only allowed values."""
+    for readme_file in sorted(Path(ROOT_PATH).rglob("README.md")):
+        metadata = update_projects_table.extract_metadata(readme_file)
+
+        if metadata:
```
daabr: Multiple levels of indentation are sometimes necessary, but it's better to flatten the code when it makes sense:
1. Instead of `if metadata`:
   ```py
   if not metadata:
     continue
   ```
2. Instead of a multiline error message:
   ```py
   error = f"..."
   assert condition, error
   ```
   (Like you did in `test_required_categorie()`)

Arguably 1 is more idiomatic in Go than in Python, and 2 is unnecessary in flatter code (e.g. in `test_metadata_block_exists`), so consider at least one of them, if not both, unless other comments make this code flatter anyway. Up to you.

tests/convention_test.py
```diff
@@ -0,0 +1,74 @@
...
+METADATA_VALID = [
+    (metadata_definitions.CATEGORIES, "categories"),
+    (metadata_definitions.INTEGRATIONS, "integrations"),
+]
+
+
+def test_metadata_block_exists():
+    """Checks for required metadata in README files."""
+    for readme_file in sorted(Path(ROOT_PATH).rglob("README.md")):
+        metadata = update_projects_table.extract_metadata(readme_file)
```
daabr: Notice that all your tests (unsurprisingly) start with the same code:
```py
    for readme_file in sorted(Path(ROOT_PATH).rglob("README.md")):
        metadata = update_projects_table.extract_metadata(readme_file)
```
You can create a helper function that returns (or even better - yields) `readme_file` and `metadata`.

...And you can also skip files that don't have metadata.

tests/convention_test.py
```diff
@@ -0,0 +1,74 @@
...
+        assert metadata["integrations"], (
+            f"{metadata['title']} metadata must have at least one integration."
+        )
+        assert metadata["categories"], (
+            f"{metadata['title']} metadata must have at least one category."
+        )
+
+
+@pytest.mark.parametrize(("allow_list", "metadata_type"), METADATA_VALID)
+def test_metadata_valid(allow_list, metadata_type):
```
daabr: Don't be afraid to use long descriptive names for test functions - it's a form of self-documentation that can make the docstring unnecessary. For example:
- `test_metadtata_contains_at_least_1_integration`
- `test_metadata_contains_all_fields`
Note that when a test fails Pytest prints the function's name, not the function's docstring.

tests/convention_test.py
```diff
@@ -0,0 +1,74 @@
...
+        )
+        assert metadata["integrations"], (
+            f"{metadata['title']} metadata must have at least one integration."
+        )
+        assert metadata["categories"], (
+            f"{metadata['title']} metadata must have at least one category."
+        )
+
+
+@pytest.mark.parametrize(("allow_list", "metadata_type"), METADATA_VALID)
```
daabr: Especially in tests, the ease of understanding the code is important. So move `METADATA_VALID` above this function, instead of the top of this file (i.e. near the only place that uses it).

Same for `TEST_CATEGORIES` in `test_required_categories()`.

tests/convention_test.py
```diff
@@ -0,0 +1,74 @@
...
+        metadata = update_projects_table.extract_metadata(readme_file)
+
+        if metadata:
+            for item in metadata[metadata_type]:
+                assert item in allow_list, (
+                    f"Invalid '{item}' in {metadata_type} in {metadata['title']}."
+                )
+
+
+@pytest.mark.parametrize(("folder_path", "categorie"), TEST_CATEGORIES)
```
daabr: Typo: single "category", plural "categories".

tests/convention_test.py
```diff
@@ -0,0 +1,74 @@
...
+            message += f" is missing the {categorie} categorie."
+            assert categorie in metadata["categories"], message
+
+
+def test_sample_name():
+    for readme_file in sorted(Path(ROOT_PATH / "samples").rglob("README.md")):
+        metadata = update_projects_table.extract_metadata(readme_file)
+
+        if metadata:
+            assert "sample" in metadata["title"], (
```
daabr:
```suggestion
            assert metadata["title"].endswith(" sample"), (
```

daabr (commented):
Given the issue ID in the branch name, please add manually a link to this PR in INT-183 (although it's not a bad thing that it's also mentioned in INT-182).

tests/convention_test.py
mario99logic: I implemented the change, but now I’m deciding whether to keep it as a constant list and use it with:
```
 @pytest.mark.parametrize(("folder_path", "categorie"), TEST_CATEGORIES)
```

or do I replace it with 
```
@pytest.mark.parametrize(
    ("folder_path", "categorie"),
    [
        (Path(ROOT_PATH / "devops"), "DevOps"),
        (Path(ROOT_PATH / "reliability"), "Reliability"),
        (Path(ROOT_PATH / "samples"), "Samples"),
    ]
)
```

daabr: either of them is fine, whichever you prefer

tests/metadata_definitions.py
```diff
@@ -0,0 +1,35 @@
...
+    "confluence",
+    "docs",
+    "drive",
+    "forms",
+    "github",
+    "githubcopilot",
+    "gmail",
+    "googlegemini",
+    "hubspot",
+    "http",
```
daabr:
```suggestion
    "http",
    "hubspot",
```

tests/metadata_definitions.py
```diff
@@ -0,0 +1,35 @@
+METADATA = ["title", "description", "integrations", "categories"]
```
daabr:
```suggestion
METADATA = ("title", "description", "integrations", "categories")
```

tests/metadata_definitions.py
```diff
@@ -0,0 +1,35 @@
...
+
+ALLOWED_INTEGRATIONS = (
+    "asana",
+    "auth0",
+    "aws",
```
daabr:
```suggestion
    "auth0",
    "autokitteh",
    "aws",
```

tests/convention_test.py
```diff
@@ -0,0 +1,76 @@
...
+    (Path(ROOT_PATH / "devops"), "DevOps"),
+    (Path(ROOT_PATH / "reliability"), "Reliability"),
+    (Path(ROOT_PATH / "samples"), "Samples"),
+]
+
+
+@pytest.mark.parametrize(("folder_path", "category"), TEST_CATEGORIES)
+def test_projects_include_required_category(folder_path, category):
+    for readme_file, metadata in get_metadata_files(folder_path):
+        error = f"Project is missing the {category} category: {readme_file}"
```
daabr:
```suggestion
        error = f"Metadata is missing the {category!r} category: {readme_file}"
```

daabr (commented):
Only a few nits. I'd approve, but I know that some of these tests are currently failing. Any idea why the CI doesn't show this?

daabr (approved):

PR #200, opened 2025-02-07
demo: incidenter

itayd:


incidenter/README.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Incident management automation
+description: Slack/Zoom/Height Incident Automation
+integrations: ["slack", "zoom", "height"]
+categories: ["DevOps"]
```
daabr:
```suggestion
categories: ["Reliability"]
```

incidenter/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+# Incident Management Automation
+
+An incident can be created by mentioning the Slack bot like so:
+
+```
+@ak incident title
+```
+
+This will create a Zoom meeting, a Height ticket and a Slack channel for the incident.
```
daabr:
```suggestion
This will create a Zoom meeting, a Height ticket, and a Slack channel for the incident.
```

incidenter/README.md
```diff
@@ -0,0 +1,18 @@
...
+In the Slack channel, incident is resolved by writing "!resolve why this was resolved". A note can be added
```
daabr, line 17: A question, not a comment: any reason you preferred to use "!command" messages as triggers instead of slash commands?

incidenter/autokitteh.yaml
```diff
@@ -0,0 +1,27 @@
+# This YAML file is a declarative manifest that describes the setup of
+# an AutoKitteh project that monitors comments on GitHub issues.
```
daabr: Copy-paste error
```suggestion
# an AutoKitteh project that automates incident management.
```

incidenter/height.py
```diff
@@ -0,0 +1,41 @@
+from os import getenv
```
daabr, line 1: Prefer importing modules, not classes and functions, unless there's a readability advantage
```suggestion
import os
```

incidenter/height.py
```diff
@@ -0,0 +1,41 @@
...
+
+from autokitteh.height import height_client
+
+
+height = height_client("height")
+
+
+LIST_ID = getenv("HEIGHT_LIST_ID")
```
daabr: Ensure LIST_ID is never `None`
```suggestion
LIST_ID = os.getenv("HEIGHT_LIST_ID", "")
```

incidenter/height.py
```diff
@@ -0,0 +1,41 @@
+from os import getenv
+
+from autokitteh.height import height_client
+
+
+height = height_client("height")
```
daabr: vars should be below consts, i.e. below `LIST_ID`

incidenter/height.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+LIST_ID = getenv("HEIGHT_LIST_ID")
+
+_root_url = "https://api.height.app/"
```
daabr: This should be a const? i.e. `_ROOT_URL`

incidenter/height.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def _post(path: str, data: dict) -> dict:
+    resp = height.post(f"{_root_url}{path}", json=data)
```
daabr: https://docs.python.org/2/library/urlparse.html#urlparse.urljoin

incidenter/program.py
```diff
@@ -0,0 +1,145 @@
+from os import getenv
+import re
+
+from autokitteh import next_event, subscribe
+from autokitteh.slack import slack_client
+
+from height import add_task_message
+from height import create_task
```
daabr: Prefer importing modules, not functions. `height.create_task` is clearer to the reader as referring to Height.

incidenter/program.py
```diff
@@ -0,0 +1,145 @@
...
+CHANNEL_PREFIX = getenv("CHANNEL_PREFIX")
+
+slack_client = slack_client("slack")
+
+pattern = re.compile(r"<.+> incident *(.*)")
+
+
+def on_slack_app_mention(event):
+    m = pattern.match(event.data.text)
+    if not m:
```
daabr, line 21: It seems like a better idea to put this regex filter in the trigger, but it's enough to put a TODO reminder if you don't want to do it now

incidenter/program.py
```diff
@@ -0,0 +1,145 @@
...
+from height import add_task_message
+from height import create_task
+from zoom import create_meeting
+
+
+CHANNEL_PREFIX = getenv("CHANNEL_PREFIX")
+
+slack_client = slack_client("slack")
+
+pattern = re.compile(r"<.+> incident *(.*)")
```
daabr: 1. Start at the beginning of the string
2. Don't be fooled by `">"` later in the string
3. `"\s*"` is more clear than `" *"`
```suggestion
pattern = re.compile(r"^<.+?> incident\s*(.*)")
```

incidenter/program.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def on_slack_app_mention(event):
+    m = pattern.match(event.data.text)
+    if not m:
+        print("irrelevant")
+        return
+
+    incident = _start(event.data.user, event.data.channel, event.data.ts, m.groups()[0])
```
daabr: Better:
```suggestion
    incident = _start(event.data.user, event.data.channel, event.data.ts, m.group(1))
```

incidenter/program.py
```diff
@@ -0,0 +1,145 @@
...
+
+    _track(incident)
+
+
+def _start(
+    slack_user: str, trigger_channel_id: str, trigger_ts: str, title: str
+) -> dict:
+    task = create_task(title, "New incident created from slack", "inProgress")
```
daabr:
```suggestion
    task = create_task(title, "New incident created from Slack", "inProgress")
```

incidenter/program.py
```diff
@@ -0,0 +1,145 @@
...
+        text, user = evt.text.strip(), evt["user"]
+
+        if not text.startswith("!"):
+            continue
+
+        parts = text[1:].strip().split(" ", 1)
+        cmd = parts[0]
+        rest = parts[1] if len(parts) > 1 else ""
+
+        if cmd == "abandon":
```
daabr: https://docs.python.org/3/tutorial/controlflow.html#tut-match

incidenter/program.py
```diff
@@ -0,0 +1,145 @@
...
+
+            slack_client.reactions_add(
+                channel=evt["channel"],
+                timestamp=evt["ts"],
+                name="memo",
+            )
+
+
+def _resolve(incident: dict, slack_user: str, msg: str):
+    text = "🎉 Incident resolved!"
```
daabr, line 121: Optional nit
```suggestion
    text = ":tada: Incident resolved!"
```

incidenter/program.py
```diff
@@ -0,0 +1,145 @@
...
+
+    zoom_url = create_meeting(f"Incident: {title}")
+
+    channel = slack_client.conversations_create(
+        name=f"{CHANNEL_PREFIX}{task["index"]}", is_private=False
+    ).get("channel")
+
+    slack_client.conversations_setTopic(
+        channel=channel["id"],
+        topic=f"⚠ Incident: {title}",
```
daabr, line 43: Optional nit (not relevant in line 59 because it would be cumbersome there)
```suggestion
        topic=f":warning: Incident: {title}",
```

daabr (approved):

daabr (approved):
Lots of comments, but they're all minor, so approving in advance. Need to fix the lint failure.

incidenter/program.py
itayd: tbh i think it's better as it is now...

incidenter/README.md
itayd: i'd rather it be seen by everybody, better transparency and can be invoked at an appropriate context.

incidenter/height.py
itayd: i actually like how it is now - when you see a `getenv`, adding the `os.` adds nothing.

incidenter/height.py
itayd: why does it matter? None or "" in json serializing further ahead will yield an error anyway.

itayd: added a check

incidenter/height.py
itayd: done

incidenter/height.py
itayd: done

incidenter/program.py
itayd: here i tend to agree

incidenter/program.py
itayd: this makes sense... will check

itayd: I added a note re that.

incidenter/program.py
itayd: done

incidenter/program.py
itayd: cool, done

PR #212, opened 2025-03-03
refactor(incidenter): style cleanups

daabr:
Readme still isn't conformant to our usual template, but fixing that isn't urgent for now.


incidenter/height.py
```diff
@@ -1,18 +1,17 @@
...
 from autokitteh.height import height_client
-from urlparse import urljoin
 
 
-_LIST_ID = getenv("HEIGHT_LIST_ID")
-
 _ROOT_URL = "https://api.height.app/"
 
-height = height_client("height")
+_HEIGHT_LIST_ID = os.getenv("HEIGHT_LIST_ID")
```
pashafateev, line 9: why underscore at the beginning?

incidenter/height.py
```diff
@@ -1,18 +1,17 @@
...
 
+if not _HEIGHT_LIST_ID:
+    raise ValueError("HEIGHT_LIST_ID project variable must be set")
```
pashafateev, line 12: no underscore here

incidenter/program.py
```diff
@@ -8,16 +8,14 @@
...
-slack_client = slack_client("slack")
-
-pattern = re.compile(r"^<.+?>\s*incident\s*(.*)")
+slack_client = slack_client("slack_conn")
 
 
 def on_slack_app_mention(event):
     # Another option is to put this straight in the trigger definition as "filter".
-    m = pattern.match(event.data.text)
+    m = re.compile(r"^<.+?>\s*incident\s*(.*)").match(event.data.text)
```
pashafateev, line 18: If we can avoid regex in the workflow code I think it's wise.

incidenter/height.py
daabr: A "_" prefix is a "soft" Python convention for denoting something as private. "Soft" in the sense that it's not enforced by the language, in contrast to "__" (which does make it harder and uglier to import it across modules).

incidenter/height.py
daabr: This is the AK project var name, not the Python const name.

incidenter/program.py
daabr: I agree, but I just moved it from the module's global scope into the function. I don't want to modify this at this time.

pashafateev (approved):

PR #214, opened 2025-03-05
fix: Auth0 to HubSpot project fix 

mario99logic:
after I fixed the error in the ticket I got another error related to returned data from Auth0 not containing the correct fields , which I also fixed

Refs: INT-317

### DIFF
--- a/auth0_to_hubspot/program.py
+++ b/auth0_to_hubspot/program.py
@@ -63,7 +63,7 @@ def _create_hubspot_contact(user):
     return SimplePublicObjectInput(properties=user_data)
 
 
-def extract_name(user):
+def _extract_name(user):
     """Extracts first and last name from user."""
     if "given_name" in user and "family_name" in user:
         return user["given_name"], user["family_name"]
--- a/auth0_to_hubspot/program.py
+++ b/auth0_to_hubspot/program.py
@@ -53,8 +53,7 @@ def add_new_users(users):
 
 def _create_hubspot_contact(user):
     """Convert Auth0 user data to HubSpot contact format."""
-    first_name = user.get("given_name") or user.get("name", "").split(" ")[0]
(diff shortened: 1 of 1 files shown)


auth0_to_hubspot/program.py
```diff
@@ -50,9 +53,12 @@ def add_new_users(users):
 
 def _create_hubspot_contact(user):
     """Convert Auth0 user data to HubSpot contact format."""
+    first_name = user.get("given_name") or user.get("name", "").split(" ")[0]
+    last_name = user.get("family_name") or " ".join(user.get("name", "").split(" ")[1:])
```
pashafateev: This is difficult to understand for a simple project. I understand why you're doing this. Maybe add a separate function and call it `extract_name` with two return types or two separate functions. Hide the complexity and maybe add a comment in there.

pashafateev:
have you tested your solution?

auth0_to_hubspot/program.py
```diff
@@ -50,9 +53,22 @@ def add_new_users(users):
...
         "email": user["email"],
-        "firstname": user["given_name"],
-        "lastname": user["family_name"],
+        "firstname": first_name,
+        "lastname": last_name,
     }
     return SimplePublicObjectInput(properties=user_data)
+
+
+def extract_name(user):
```
daabr:
```suggestion
def _extract_name(user):
```

auth0_to_hubspot/program.py
```diff
@@ -50,9 +53,22 @@ def add_new_users(users):
 
 def _create_hubspot_contact(user):
     """Convert Auth0 user data to HubSpot contact format."""
+    first_name, last_name = extract_name(user)
```
daabr:
```suggestion
    first_name, last_name = _extract_name(user)
```

daabr (approved):

PR #217, opened 2025-03-13
openai agent demo

itayd:
Allows to use chat with custom openai agents.

### DIFF
--- a/ai_agents/openai_agent_researcher/README.md
+++ b/ai_agents/openai_agent_researcher/README.md
@@ -1,7 +1,7 @@
 ---
 title: OpenAI Agent Researcher
 description: A Slack-based research agent workflow.
-integrations: ["chatgpt", "slack", "openai"]
+integrations: ["chatgpt", "slack"]
 categories: ["AI"]
 ---
 
@@ -9,47 +9,11 @@ categories: ["AI"]
 
 An interactive AI research assistant that uses OpenAI Agents to perform web searches and gather information based on user queries. Built on the durable AutoKitteh platform, this assistant communicates with users through Slack, plans research tasks, and delivers comprehensive reports with fault-tolerant execution.
 
-## Features
+API documentation:
 
-- **Interactive Planning**: Generates a research plan based on user queries
-- **Web Search Integration**: Searches the web for relevant information
-- **Human Collaboration**: Can ask specific people questions as part of the research
-- **Report Generation**: Synthesizes research findings into comprehensive reports
-- **Slack Integration**: Communicates with users through Slack
+- openai-agents: https://platform.openai.com/docs/guides/agents
 
-## Components
-
-### Core Modules
-
-- **workflow.py**: Main entry point handling the end-to-end research workflow
-- **ai.py**: Defines AI agents for planning, searching, and reporting
-- **data.py**: Contains data models for the research plan and report
-- **slack.py**: Handles Slack communication
-- **tools.py**: Defines tools for the OpenAI agents
-
-### Agents
-
-1. **PlannerAgent**: Creates a research plan from user queries
-2. **SearchAgent**: Performs web searches and summarizes results
-3. **ReporterAgent**: Synthesizes findings into a cohesive report
-
-## Setup
-
-### Prerequisites
-
-- OpenAI API key with access to GPT-4
-- Slack workspace with bot permissions
-- Access to AutoKitteh Cloud
-
-### Setup
-
-1. In the AutoKitteh Cloud UI:
-   - Configure the `OPENAI_API_KEY` project variable with your API key
-   - Configure the `INVOCATION_CMD` variable to set the command prefix (default: "research")
-2. Initialize the Slack connection through the AutoKitteh Cloud UI
-3. Deploy the project using AutoKitteh Cloud
-
-## Usage
+## How It Works
 
 1. In any Slack channel where the bot is present, use the command:
 
@@ -67,8 +31,24 @@ An interactive AI research assistant that uses OpenAI Agents to perform web sear
 
 3. The assistant can be instructed to send reports to specific users with the appropriate command.
 
-## Workflow
+## Cloud Usage
 
-1. **Planning Phase**: The PlannerAgent creates a research plan with 3-10 tasks
-2. **Execution Phase**: Tasks are executed (searches, questions to individuals)
-3. **Reporting Phase**: Results are synthesized into a report
+1. Initialize Slack connection
+2. Configure the `OPENAI_API_KEY` project variable with your API key
+3. Configure the `INVOCATION_CMD` variable to set the command prefix (default: "research")
+4. Deploy project
+
+## Trigger Workflow
+
+> [!IMPORTANT]
+> Ensure the Slack connection is initialized; otherwise, the workflow will raise a `ConnectionInitError`.
+
+In any Slack channel where the bot is present, use the command:
+
+```
+!research [your research question]
+```
+
+## Self-Hosted Deployment
+
+Follow [these detailed instructions](https://docs.autokitteh.com/get_started/deployment) to deploy the project on a self-hosted server.
--- a/ai_agents/openai_agent_researcher/README.md
+++ b/ai_agents/openai_agent_researcher/README.md
@@ -1,0 +1,74 @@
+---
+title: OpenAI Agent Researcher
+description: A Slack-based research agent workflow.
+integrations: ["chatgpt", "slack", "openai"]
+categories: ["AI"]
+---
+
+# OpenAI Agent Researcher
+
+An interactive AI research assistant that uses OpenAI Agents to perform web searches and gather information based on user queries. Built on the durable AutoKitteh platform, this assistant communicates with users through Slack, plans research tasks, and delivers comprehensive reports with fault-tolerant execution.
+
+## Features
+
+- **Interactive Planning**: Generates a research plan based on user queries
+- **Web Search Integration**: Searches the web for relevant information
+- **Human Collaboration**: Can ask specific people questions as part of the research
+- **Report Generation**: Synthesizes research findings into comprehensive reports
+- **Slack Integration**: Communicates with users through Slack
+
+## Components
+
+### Core Modules
+
+- **workflow.py**: Main entry point handling the end-to-end research workflow
+- **ai.py**: Defines AI agents for planning, searching, and reporting
+- **data.py**: Contains data models for the research plan and report
+- **slack.py**: Handles Slack communication
+- **tools.py**: Defines tools for the OpenAI agents
+
+### Agents
+
+1. **PlannerAgent**: Creates a research plan from user queries
+2. **SearchAgent**: Performs web searches and summarizes results
+3. **ReporterAgent**: Synthesizes findings into a cohesive report
+
+## Setup
+
+### Prerequisites
+
+- OpenAI API key with access to GPT-4
+- Slack workspace with bot permissions
+- Access to AutoKitteh Cloud
+
+### Setup
+
+1. In the AutoKitteh Cloud UI:
+   - Configure the `OPENAI_API_KEY` project variable with your API key
+   - Configure the `INVOCATION_CMD` variable to set the command prefix (default: "research")
+2. Initialize the Slack connection through the AutoKitteh Cloud UI
+3. Deploy the project using AutoKitteh Cloud
+
+## Usage
+
+1. In any Slack channel where the bot is present, use the command:
+
+   ```
+   !research [your research question]
+   ```
+
+2. The assistant will:
+
+   - Generate a research plan
+   - Ask for confirmation
+   - Execute searches
+   - (Optionally) Ask specific people questions
+   - Compile a final report
+
+3. The assistant can be instructed to send reports to specific users with the appropriate command.
+
+## Workflow
+
+1. **Planning Phase**: The PlannerAgent creates a research plan with 3-10 tasks
+2. **Execution Phase**: Tasks are executed (searches, questions to individuals)
+3. **Reporting Phase**: Results are synthesized into a report
--- a/ai_agents/openai_agent_researcher/ai.py
+++ b/ai_agents/openai_agent_researcher/ai.py
@@ -14,9 +14,9 @@ from openai import RateLimitError
 from data import Report
 from data import ResearchPlan
 from data import SearchResearchItem
-from tools import send_slack_report
 from slack import next_input
 from slack import send
+from tools import send_slack_report
 
 
 _plan_agent = Agent(
--- a/ai_agents/openai_agent_researcher/ai.py
+++ b/ai_agents/openai_agent_researcher/ai.py
@@ -30,7 +30,7 @@ A task can be either:
   If a user explicitly specifies a time limit for a specific user, set it as such.
   Do this only if the user explicitly specifies a person to ask.
 For each task result, if applicable, default max tokens to None, unless user explicitly
-specified otherwise. User cannot be allowed to specifiy max tokens below 16.
+specified otherwise. User cannot be allowed to specify max tokens below 16.
 You can also modify an existing plan, by adding or removing searches.
 Always provide the complete plan as output along with an indication if the user
 considers it final.
@@ -45,7 +45,7 @@ _search_agent = Agent(
     instructions="""
 You are a research assistant. Given a search term, you search the web for that term and
 produce a concise summary of the results. The summary must 2-3 paragraphs and less than
-300 words. Capture the main points. Write succintly, no need to have complete sentences
+300 words. Capture the main points. Write succinctly, no need to have complete sentences
 or good grammar. This will be consumed by someone synthesizing a report, so its vital
 you capture the essence and ignore any fluff. Do not include any additional commentary
 other than the summary itself.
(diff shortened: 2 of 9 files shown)


ai_agents/openai_agent_researcher/workflow.py
```diff
@@ -0,0 +1,72 @@
...
+from os import getenv
+
+import data
+
+from ai import plan
+from ai import report
+from ai import search
+import slack
+from slack import ask
+from slack import send
```
daabr: It's a bad practice to import specific functions, and it's much more readable to see where they're coming from when they're being called.

There are exceptions to that, especially with stdlib and 3rd-party imports, but not this.

```suggestion
import ai
import data
import slack
```

ai_agents/openai_agent_researcher/workflow.py
```diff
@@ -0,0 +1,72 @@
...
+    There are three phases:
+    1. Plan the search.
+    2. Execute the search.
+    3. Report the results.
+    """
+    # Plan the search.
+    search_plan = plan(q)
+
+    send("Now I will execute on the plan.\n")
+
```
daabr: The empty lines here are making it hard to understand this block, not easier.
```suggestion
    send("Now I will execute on the plan.\n")
```

ai_agents/openai_agent_researcher/workflow.py
```diff
@@ -0,0 +1,72 @@
...
+                    tasks[f"According to the user {who['real_name']}"] = answer
+                else:
+                    tasks[f"According to the user {t.who}"] = (
+                        f"could not figure out which user is {t.who}"
+                    )
+
+    # Summarize and report the results.
+    send("All tasks complete, summarizing results...")
+
```
daabr:
```suggestion
```

ai_agents/openai_agent_researcher/workflow.py
```diff
@@ -0,0 +1,72 @@
...
+    if not text.startswith(_prefix):
+        print("irrelevant")
+        return
+
+    slack.init(event.data.channel, event.data.ts)
+
+    q = text.removeprefix(_prefix)
+
+    print(f"Q: {q}")
+
```
daabr:
```suggestion
    print(f"Q: {q}")
```

ai_agents/openai_agent_researcher/slack.py
daabr, line 1: Even though "slack" != "slack_sdk", it's clearer and safer to use unique names for local modules, like "slack_helper", "slack_util", etc.

ai_agents/openai_agent_researcher/data.py
```diff
@@ -0,0 +1,108 @@
...
+
+{"Final" if self.is_final else "Draft"} Plan:
+"""
+        for task in self.tasks:
+            text += f"{task}\n"
+
+        return text
+
+
+class Report(BaseModel):
```
daabr, line 93: Why not https://docs.python.org/3/library/dataclasses.html ?

ai_agents/openai_agent_researcher/tools.py
```diff
@@ -0,0 +1,12 @@
...
+
+from agents import function_tool
+from data import Report
+
+from slack import send
+
+
+@function_tool
+async def send_slack_report(r: Report, user: str) -> None:
+    """Send a slack report to the given user."""
```
daabr:
```suggestion
    """Send a Slack report to the given user."""
```

ai_agents/openai_agent_researcher/tools.py
```diff
@@ -0,0 +1,12 @@
+"""Agent tools for the OpenAI agent."""
+
+from agents import function_tool
+from data import Report
+
+from slack import send
```
daabr: `agents` is a third-party import, `data` and `slack` are not.

ai_agents/openai_agent_researcher/data.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __str__(self):
+        return f"""{self.common}Search Query: "{self.query}"
+{f"Max Tokens: {self.max_tokens}" if self.max_tokens else ""}"""
+
+
+class AskSomeoneResearchItem(BaseModel):
+    """Ask someone for additional information."""
+
+    wait_time_in_seconds: int = None
```
daabr: You're doing this in a lot of places: `var: type = None` is a typing bug.

Either the type hint is `type | None` or the default is an instance of the specified type.

Note that `0` and `""` are also evaluated as false like `None`, so you don't need `None`.

ai_agents/openai_agent_researcher/data.py
```diff
@@ -0,0 +1,108 @@
...
+
+    query: str
+    "The search term to use for the web search."
+
+    common: ResearchItemCommon
+    "Common fields for research items"
+
+    def __str__(self):
+        return f"""{self.common}Search Query: "{self.query}"
+{f"Max Tokens: {self.max_tokens}" if self.max_tokens else ""}"""
```
daabr, line 37: A multiline string here is confusing, especially since you only need a single `\n`.

ai_agents/openai_agent_researcher/data.py
```diff
@@ -0,0 +1,108 @@
...
+
+    common: ResearchItemCommon
+    "Common fields for research items"
+
+    def __str__(self):
+        time_limit = (
+            f"Time Limit: {self.wait_time_in_seconds} seconds"
+            if self.wait_time_in_seconds
+            else ""
+        )
```
daabr: Ternary conditionals should be a single line, if they're wrapped you should use a regular `if-else`.

ai_agents/openai_agent_researcher/data.py
```diff
@@ -0,0 +1,108 @@
...
+
+        return f"""{self.common}Person: {self.who}
```
daabr, line 60: Same as `SearchResearchItem.__str__`

ai_agents/openai_agent_researcher/slack.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+slack = slack_client("slack_conn")
+
+_ch = None
+_ts = None
+_subscription = None
+
+
+def init(ch, ts):
```
daabr: Python is not Go. Readable Python code uses clear var names, not abbreviations. Also, if  other functions in this project use type hints, why not here?

ai_agents/openai_agent_researcher/slack.py
```diff
@@ -0,0 +1,100 @@
...
+    _subscription = subscribe(
+        "slack_conn",
+        f"data.type == 'message' && data.bot_id == '' && data.thread_ts == '{ts}'",
+    )
+
+    _ch = ch
+    _ts = ts
+
+
+def _lookup_user(who: str) -> any:
```
daabr: `any` here is both incorrect and useless, it should be `str`

ai_agents/openai_agent_researcher/slack.py
```diff
@@ -0,0 +1,100 @@
...
+    elif who.startswith("@"):
+        user_id = who[1:]
+    elif "@" in who:
+        try:
+            user_id = slack.users_lookupByEmail(email=who)["user"]["id"]
+        except SlackApiError as e:
+            print(f"error: {e}")
+            return None
+    else:
+        return None
```
daabr, line 41: Returning `None` forces the callers to handle that case. Isn't it more logical to just raise an exception? That's what happens anyway if the call below fails. Or at least an empty string as a signal for invalid input?

ai_agents/openai_agent_researcher/slack.py
```diff
@@ -0,0 +1,100 @@
...
+
+    user = slack.users_info(user=user_id).get("user")
```
daabr: If this call fails it raises an exception. Is that OK, or do you want to return `""`?

ai_agents/openai_agent_researcher/slack.py
```diff
@@ -0,0 +1,100 @@
...
+
+    print(f"lookup: {who} -> {user}")
+
+    return user
+
+
+def _post(text, user_id: str = None):
```
daabr:
```suggestion
def _post(text: str, user_id: str = ""):
```

ai_agents/openai_agent_researcher/slack.py
```diff
@@ -0,0 +1,100 @@
...
+    ch, ts = _ch, _ts
+    if user_id:
+        ch, ts = user_id, None
+
+    print(f"{ch}{(',' + ts) if ts else ''}: {text}")
+    slack.chat_postMessage(channel=ch, text=text, thread_ts=ts)
+
+
+def send(x: any, who: str = None):
```
daabr: 1. It's entirely unclear what `x` is
2. An `str` variable can't be `None`, and you probably don't need it anyway

ai_agents/openai_agent_researcher/slack.py
```diff
@@ -0,0 +1,100 @@
...
+    event = next_event(_subscription)
+    if event is None:
+        raise EOFError
+
+    print(f"Q: {event.text}")
+
+    return event.text
+
+
+def ask(what: str, who: str, t: int = None) -> tuple[str, str]:
```
daabr: `int` can't be `None`, and probably `0` is fine too, no?

ai_agents/openai_agent_researcher/autokitteh.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - name: OPENAI_API_KEY
+      secret: true
+      # Put here your Open AI API key.
+      value: "sk-..."
+
+  connections:
+    # We use Slack to interact with the users.
+    - name: slack_conn
+      integration: slack
+    # gMail connection used to sending reports to users.
```
daabr:
```suggestion
    # Gmail connection used to sending reports to users.
```

ai_agents/openai_agent_researcher/ai.py
```diff
@@ -0,0 +1,146 @@
...
+from agents.run import RunConfig
+from autokitteh import activity
+from data import Report
+from data import ResearchPlan
+from data import SearchResearchItem
+from openai import RateLimitError
+from tools import send_slack_report
+
+from slack import next_input
+from slack import send
```
daabr: See comment about imports in `workflow.py`

ai_agents/openai_agent_researcher/ai.py
```diff
@@ -0,0 +1,146 @@
+"""AI Agents definitions."""
+
+import asyncio
+from time import sleep
```
daabr, line 4: No action, just FYI: `from time import sleep` is a good exception to the rule of thumb to import modules and not module contents, because it's just a single member of the module, and also it's obvious in the code where `sleep()` is coming from.

ai_agents/openai_agent_researcher/README.md
```diff
@@ -0,0 +1,74 @@
...
+description: A Slack-based research agent workflow.
+integrations: ["chatgpt", "slack", "openai"]
+categories: ["AI"]
+---
+
+# OpenAI Agent Researcher
+
+An interactive AI research assistant that uses OpenAI Agents to perform web searches and gather information based on user queries. Built on the durable AutoKitteh platform, this assistant communicates with users through Slack, plans research tasks, and delivers comprehensive reports with fault-tolerant execution.
+
+## Features
```
daabr: FYI: Kittehub uses this template for readmes: https://github.com/autokitteh/catcafe/blob/main/kittehub/README_template.md

We're not too strict about it, but it looks like you're using something else as a template.

ai_agents/openai_agent_researcher/data.py
itayd: according to openai-agents docs, that's what they want.

ai_agents/openai_agent_researcher/tools.py
itayd: done

ai_agents/openai_agent_researcher/data.py
itayd: removed the None, will fix elsewhere as well

ai_agents/openai_agent_researcher/data.py
itayd: would've needed to break it down anyway since it's too long

ai_agents/openai_agent_researcher/data.py
itayd: done

ai_agents/openai_agent_researcher/data.py
itayd: line would've been too long

ai_agents/openai_agent_researcher/slack.py
itayd: this is clear enough tbh

itayd: added type hints though

ai_agents/openai_agent_researcher/slack.py
itayd: it's actually a dict[str, any].

fixed.

ai_agents/openai_agent_researcher/slack.py
itayd: we should never have an exception here. the user must always exist.

ai_agents/openai_agent_researcher/slack.py
itayd: it might also be none

ai_agents/openai_agent_researcher/slack.py
itayd: fixed

ai_agents/openai_agent_researcher/slack.py
itayd: can be None, fixed

ai_agents/openai_agent_researcher/autokitteh.yaml
itayd: removed gmail, was left by mistake

ai_agents/openai_agent_researcher/README.md
itayd: yeah i used ai to generate it and it did a great job, so i'd like to keep it

ai_agents/openai_agent_researcher/slack.py
itayd: i generally agree with you about the exception, but here this really simplifies stuff as the caller doesn't really care why - it just need to know it didn't work.

and None is good as it doesn't mask the fact that there have been an error.

ai_agents/openai_agent_researcher/tools.py
itayd: yeah well ruff is messing with me

ai_agents/openai_agent_researcher/slack.py
itayd: i think this is too much.

ai_agents/openai_agent_researcher/workflow.py
itayd: done

ai_agents/openai_agent_researcher/workflow.py
itayd: done

ai_agents/openai_agent_researcher/workflow.py
itayd: fixed

ai_agents/openai_agent_researcher/workflow.py
itayd: done

ai_agents/openai_agent_researcher/ai.py
itayd: done

ai_agents/openai_agent_researcher/tools.py
daabr: probably because of this: https://github.com/autokitteh/kittehub/pull/217#discussion_r2008345762

ai_agents/openai_agent_researcher/slack.py
daabr: that's not something you can guarantee, so you should catch this error

ai_agents/openai_agent_researcher/slack.py
daabr: not according to your typing

ai_agents/openai_agent_researcher/README.md
daabr: If it did a great job I wouldn't have to make this comment. The text is fine, the section layout is not. Keeping it as-is will only force someone else to rewrite it later.

ai_agents/openai_agent_researcher/tools.py
itayd: it's not, it's because i don't have an __init__.py file so i can't do `import .whatever`.

ai_agents/openai_agent_researcher/slack.py
itayd: yeah, fixed.

ai_agents/openai_agent_researcher/README.md
itayd: done

ai_agents/openai_agent_researcher/slack.py
itayd: done

daabr (approved):

PR #220, opened 2025-03-17
feat: add "send email" walkthrough

pashafateev:
This walkthrough expects data; can the parameters be pre-populated?
image

```
{
    "recipient": "[ENTER RECIPIENT EMAIL HERE]",
    "sender": "[ENTER YOUR EMAIL HERE]",
    "subject": "[ENTER SUBJECT HERE]",
    "body": "[ENTER EMAIL BODY HERE]"
}
```

The connection isn't configured. This is intentional so the user will have to create the connection and then use the connection name in the code.

tested: successfully sent and received email

### DIFF
--- a/walkthroughs/send_email/program.py
+++ b/walkthroughs/send_email/program.py
@@ -6,14 +6,18 @@ from autokitteh.google import gmail_client
 from googleapiclient.errors import HttpError
 
 
-def on_manual_run(event):
-    gmail = gmail_client("<INSERT_CONNECTION_NAME>").users()
+gmail = gmail_client("<INSERT_CONNECTION_NAME>").users()
 
-    msg = f"""From: {event.data.sender}
-    To: {event.data.recipient}
-    Subject: Meow! {event.data.subject}
 
-    {event.data.body}"""
+def on_manual_run(_):
+    profile = gmail.getProfile(userId="me").execute()
+
+    msg = f"""From: {profile["emailAddress"]}
+    To: {profile["emailAddress"]}
+    Subject: Test from AutoKitteh
+
+    Meow! Just wanted to let you know that your cat overlords are pleased with your
+    service today. Keep up the good work with the treats and belly rubs!"""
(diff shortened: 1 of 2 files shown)


walkthroughs/send_email/program.py
```diff
@@ -0,0 +1,26 @@
...
+import base64
+
+from autokitteh.google import gmail_client
+from googleapiclient.errors import HttpError
+
+
+def on_manual_run(event):
+    gmail = gmail_client("<INSERT_CONNECTION_NAME>").users()
+
+    msg = f"""From: {event.data.sender}
```
haimzlato: Is it a must to have this? the default will not be the user?

walkthroughs/send_email/program.py
```diff
@@ -0,0 +1,26 @@
...
+    To: {event.data.recipient}
+    Subject: Meow! {event.data.subject}
+
+    {event.data.body}"""
```
haimzlato: I suggest just to put Text, not relay on the event to come.
As simple as possible.

walkthroughs/send_email/program.py
pashafateev: Done

walkthroughs/send_email/program.py
pashafateev: Done

walkthroughs/send_email/program.py
```diff
@@ -0,0 +1,30 @@
+"""Simple workflow for sending emails."""
+
+import base64
+
+from autokitteh.google import gmail_client
+from googleapiclient.errors import HttpError
```
haimzlato, line 6: @pashafateev let's add a Gmail connection to the project. Then you already have the connection name. The user only needs to authenticate and that's it.

pashafateev: Done

haimzlato (approved):

PR #224, opened 2025-03-24
fix: Invoice project fixes

mario99logic:
fixed bug mentioned in ref: INT-354 , and also other errors on the way


daabr (approved):
Approving in advance, still need to fix readme + import blocks

PR #232, opened 2025-04-28
feat: add langgraph project

mario99logic:


ai_agents/langgraph_bot/program.py
```diff
@@ -0,0 +1,91 @@
...
+from langgraph.graph import END
+from langgraph.graph import START
+from langgraph.graph import StateGraph
+from langgraph.graph.message import add_messages
+from langgraph.prebuilt import ToolNode
+from langgraph.prebuilt import tools_condition
+
+
+GOOGLE_API_KEY = os.getenv("GOOGLE_API_KEY")
+TAVILY_API_KEY = os.getenv("TAVILY_API_KEY")
```
efiShtain: you don't really use this variables, why are they defined ?

efiShtain (approved):

ai_agents/langgraph_bot/program.py
mario99logic: I do use them, they are needed for the langgraph to work, they are used internally by langgraph.

PR #238, opened 2025-05-20
feat: add SFTP demo

mario99logic:


devops/sftp/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Configuration
+
+Update the `SFTP_CONFIG` dictionary in the source code to match your SFTP server details:
+
+```python
+SFTP_CONFIG = {
+    "host": "your.sftp.host",
+    "port": 22,
+    "username": "your_username",
```
itayd, line 40: do you think it might be better to set these in vars?

devops/sftp/program.py
```diff
@@ -0,0 +1,60 @@
...
+SFTP_CONFIG = {
+    "host": "test.rebex.net",
+    "port": 22,
+    "username": "demo",
+    "password": "password",
+    "remote_path": "/readme.txt",
+}
+
+
+HTTP_TARGET = os.getenv("HTTP_TARGET")
```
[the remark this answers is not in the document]
itayd: this might be a good idea?

devops/sftp/program.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def send_file_http(file_io, filename, url):
+    file_io.seek(0)
+    files = {"file": (filename, file_io)}
+    response = requests.post(url, files=files, timeout=10)
+    return response.status_code, response.text
+
+
+@autokitteh.activity
```
itayd, line 52: document while you're using activity here:
- big files boundary
- passing a stream around which is not serializable

devops/sftp/README.md
mario99logic: I think its easier and faster to set them up in the code because there are too many vars.

haimzlato (approved):

PR #242, opened 2025-06-03
feat: implement mailbox change func in gmail sample

mario99logic:


samples/google/gmail/program.py
```diff
@@ -169,4 +169,34 @@ def _messages_send(text):
...
+                    new_message_count += len(history_record["messagesAdded"])
+
+        if new_message_count > 0:
+            on_new_message()
+
+    except HttpError as e:
+        print(f"Error: {e.reason}")
+
+
+def on_new_message():
```
haimzlato: my only comment is that here you can put a function that writes nicely the from, To, Subject and body fields

haimzlato (approved):

PR #259, opened 2025-07-30
fix: proper google integration names

itayd:


tests/metadata_definitions.py
```diff
@@ -7,23 +7,22 @@
     "auth0",
     "autokitteh",
     "aws",
-    "calendar",
     "chatgpt",
     "confluence",
     "discord",
-    "docs",
```
RonenMars, line 14: `docs` are missing in the new file

efiShtain (approved):

tests/metadata_definitions.py
itayd: there is no need for it

itayd: no such integration

mario99logic (approved):

PR #266, opened 2025-10-09
Mario/int 494 telegram project

mario99logic:


telegram_ai_translator/program.py
```diff
@@ -0,0 +1,100 @@
...
+
+from prompts import DETECT_USAGE
+from prompts import get_detect_prompt
+from prompts import get_translate_prompt
+from prompts import HELP_TEXT
+from prompts import TRANSLATE_USAGE
+
+
+MODEL = "gemini-2.5-flash-lite"
+gemini = gemini_client("gemini_conn", model_name=MODEL)
```
efiShtain: what's the purpose of defining MODEL as "const" one line above the only usage?

telegram_ai_translator/program.py
```diff
@@ -0,0 +1,100 @@
...
+        return
+
+    target_lang = parts[1]
+    text_to_translate = parts[2]
+    prompt = get_translate_prompt(target_lang, text_to_translate)
+
+    try:
+        response = gemini.generate_content(prompt)
+        await telegram.send_message(chat_id=chat_id, text=response.text)
+    except (RuntimeError, ValueError) as e:
```
efiShtain: are you intentionally catching this specific errors?
are they related to gemini or telegram?

itayd: i think no need for try except at all. just let it error out and we'll see it in the console.

telegram_ai_translator/program.py
```diff
@@ -0,0 +1,100 @@
...
+        return
+
+    target_lang = parts[1]
+    text_to_translate = parts[2]
+    prompt = get_detect_prompt(target_lang, text_to_translate)
+
+    try:
+        response = gemini.generate_content(prompt)
+        await telegram.send_message(chat_id=chat_id, text=response.text)
+    except (RuntimeError, ValueError) as e:
```
itayd: same

itayd (approved):

PR #272, opened 2025-10-21
carb counter whatsapp demo

itayd:


carbcat/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+- **URL**: https://fdc.nal.usda.gov/download-datasets
+- **Dataset**: SR Legacy
+- **Content**: ~8,000 foods with detailed carbohydrate and portion information
+
+## Setup
+
+### Prerequisites
+
+- [AutoKitteh](https://docs.autokitteh.com/) installed and configured
```
haimzlato: This is for local installation or SaaS?
Remove all the installation and the rest. We are focusing on SaaS. At the end you can put a link to the repo for local intallation

carbcat/README.md
```diff
@@ -0,0 +1,134 @@
...
+- Twilio account with WhatsApp sandbox or production number
+- Anthropic API key (for Claude)
+
+### Installation
```
haimzlato, line 68: not relevant

carbcat/requirements.txt
```diff
@@ -0,0 +1,3 @@
+anthropic==0.71.0
```
haimzlato, line 1: Now that we have a connector, is that still the case?

carbcat/requirements.txt
```diff
@@ -0,0 +1,3 @@
...
+autokitteh
```
haimzlato, line 2: Why is that here? What do I miss?

carbcat/handlers.py
```diff
@@ -0,0 +1,83 @@
...
+    finally:
+        del_value(src)
+
+
+def _handle(body: str, to: str) -> None:
+    s = subscribe("whatsapp_message", filter=f"data.body.form.From.startsWith('{to}')")
+
+    def get_next() -> str:
+        print(f"waiting for next user message from {to}...")
+        evt = next_event(s)
```
haimzlato, line 64: what solves the problem you mentioned that a new message will also trigger a new workflow?

haimzlato (approved):

carbcat/README.md
itayd: yeah changed to club usage

carbcat/README.md
itayd: this describes how to setup everything?

carbcat/requirements.txt
itayd: this needs a specific version, so yes

carbcat/requirements.txt
itayd: for local dev, it has no problems in cloud

carbcat/handlers.py
itayd: using value storage. will add comments.

PR #281, opened 2025-11-03
fix: add description to project vars to scrapingbee project

mario99logic:


scrapingbee_fetch_news/program.py
```diff
@@ -6,15 +6,19 @@
 from autokitteh.slack import slack_client
 
 SB_API_KEY = os.getenv("SB_API_KEY")
-NEWS_URL = os.getenv("NEWS_URL", "https://www.theguardian.com/international/rss")
+NEWS_WEBSITE_URL = os.getenv(
```
RonenMars: I think you should put the "demo" URL in the `autokitteh.yaml` and not in the python, because the user might not understand where from the URL comes from, since the ENV Var is empty.
And here in python, you should check if `NEWS_WEBSITE_URL` doesn't exist in env-vars : raise an error.

In my eyes, it's better to include the "demo" URL in the `autokitteh.yaml` file rather than in the Python code, as users may not comprehend the source of the URL when the ENV variable is empty (for example - they see that the ENV variable is empty in the WebUI - but they still get a URL) . Additionally, in the Python code, it is essential to verify whether `NEWS_WEBSITE_URL` is absent from the environment variables; if so, an error should be raised.

scrapingbee_fetch_news/program.py
```diff
@@ -6,15 +6,19 @@
...
 
 
 def on_app_mention(event):
-    feed = client.get(NEWS_URL, params={"render_js": False, "block_resources": True})
+    feed = client.get(
+        NEWS_WEBSITE_URL, params={"render_js": False, "block_resources": True}
+    )
     if feed.status_code != 200:
         print("Feed fetch failed:", feed.status_code, feed.text[:300])
         return
```
RonenMars, line 31: When the RSS feed fetch fails, the function returns silently without notifying the user. Since they mentioned the app in Slack, they should receive feedback about the failure:

```suggestion
if feed.status_code != 200:
      slack.chat_postMessage(
          channel=SLACK_CHANNEL,
          text=f"Failed to fetch news feed: HTTP {feed.status_code}"
      )
      print("Feed fetch failed:", feed.status_code, feed.text[:300])
      return
```

scrapingbee_fetch_news/program.py
RonenMars, line 34: NIT: you might have cases you won't have `items` at all or the feed content isn't valid XML this code will crash.
Recommend you to add a Slack notification as I described above:
```
slack.chat_postMessage( No items found in the feed...)
```

RonenMars, line 48: Recommendation: wrap the `slack.chat_postMessage` with try-catch, in case it will fail (even if the API Key is wrong).

scrapingbee_fetch_news/program.py
```diff
@@ -6,15 +6,19 @@
...
+)
 SLACK_CHANNEL = os.getenv("SLACK_CHANNEL", "#news")
 
 client = ScrapingBeeClient(api_key=SB_API_KEY)
 slack = slack_client("slack_conn")
 
 
 def on_app_mention(event):
-    feed = client.get(NEWS_URL, params={"render_js": False, "block_resources": True})
+    feed = client.get(
```
RonenMars, line 22: Recommendation: add try-catch to `client.get` in case the network doesn't respond or the URL is bad and returns an error.

RonenMars (changes requested):

scrapingbee_fetch_news/program.py
mario99logic: I included it in the yaml

scrapingbee_fetch_news/program.py
mario99logic: Its okay to let the session finish in error, like other autokitteh projects

scrapingbee_fetch_news/program.py
mario99logic: same as above

RonenMars (approved):

PR #285, opened 2025-11-06
fix: reduce release size

RonenMars:
Exclude large files from dist.zip - Modified .github/workflows/dist.yml to exclude:
- carbcat/_raw.zip (13MB)
- .git/* directory

This reduces the distribution package size on the release branch.

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -15,7 +15,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: build
-        run: zip -r dist .
+        run: |
+          zip -r dist . \
+            -x "carbcat/_raw.zip" \
+            -x ".git/*"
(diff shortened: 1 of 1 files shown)


.github/workflows/dist.yml
```diff
@@ -15,7 +15,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: build
-        run: zip -r dist .
+        run: |
+          zip -r dist . \
+            -x "carbcat/_raw.zip" \
```
itayd, line 20: can you ignore all zip files?

itayd: actually, this will ruin the demo - it needs this raw file. lets discuss.

itayd: ignore all "^_*"

